PR #90, opened 2026-06-28
fix(host-bitbucket): restriction-delete idempotency + auth surface + repo-create project/error (3 S3 closures)

kraulerson:
## Summary
- `host_configure_protection` now validates the GET response shape before any POST; non-JSON / 4xx-5xx bodies surface as a clear "could not list existing restrictions" diagnostic with a body snippet, not the misleading downstream "failed to set <kind> restriction".
- DELETE failures are buffered per-id and flushed under `SOIF_DEBUG=1` so the operator sees which leftover restriction blocked creation.
- New auth path `BITBUCKET_API_TOKEN` (Bearer) takes precedence over the legacy `BITBUCKET_APP_PASSWORD` (Basic) — Atlassian is sunsetting App Passwords for Bitbucket Cloud in 2026.
- `host_create_repo` includes `project.key` in the POST payload when `BITBUCKET_PROJECT_KEY` is set (built with `jq -nc` to avoid shell-quoting bugs); payload shape is unchanged when unset so workspaces with a default project keep working.
- `host_require_cli` accepts either credential and its guidance documents the API-token path, the App-Password deprecation, and the new `BITBUCKET_PROJECT_KEY` env var.

## Findings closed
- code-host-bitbucket-3 — Existing-restriction deletes are silenced; idempotency can fail on POST due to leftover duplicates.
- code-host-bitbucket-4 — App Password is the only supported auth; Bitbucket Cloud is sunsetting App Passwords in 2026.
- code-host-bitbucket-5 — Repository create omits `project` and does not handle workspaces requiring project assignment.

## Test plan
TDD: nine new scenarios in `tests/host-drivers/bitbucket.test.sh` (T1-T2 idempotency, T3-T6 auth surface, T7-T9 project key + guidance), each authored RED first against `main`.

RED (baseline, before driver fix):
```
$ bash tests/host-drivers/bitbucket.test.sh
...
bitbucket.test.sh: host_verify_protection (org fail) PASSED
ASSERT FAIL [diagnostic names listing failure]: 'bitbucket driver: failed to set force restriction' does not contain 'could not list existing restrictions'
```

GREEN (after driver fix):
```
$ bash tests/host-drivers/bitbucket.test.sh
bitbucket.test.sh: host_name PASSED
bitbucket.test.sh: host_require_cli PASSED
bitbucket.test.sh: host_register_remote PASSED
bitbucket.test.sh: _bb_parse_origin PASSED
bitbucket.test.sh: host_configure_protection (personal) PASSED
bitbucket.test.sh: host_configure_protection (org) PASSED
bitbucket.test.sh: host_verify_protection (personal pass) PASSED
bitbucket.test.sh: host_verify_protection (personal fail) PASSED
bitbucket.test.sh: host_verify_protection (org pass) PASSED
bitbucket.test.sh: host_verify_protection (org fail) PASSED
bitbucket.test.sh: host_configure_protection (non-JSON GET surfaces failure) PASSED
bitbucket.test.sh: host_configure_protection (SOIF_DEBUG surfaces DELETE failure) PASSED
bitbucket.test.sh: API token uses Bearer (no -u) PASSED
bitbucket.test.sh: App password uses -u (legacy) PASSED
bitbucket.test.sh: host_require_cli (API token only) PASSED
bitbucket.test.sh: host_require_cli (no credential fails + mentions token) PASSED
bitbucket.test.sh: host_create_repo (project key included) PASSED
bitbucket.test.sh: host_create_repo (project key omitted) PASSED
bitbucket.test.sh: host_require_cli (mentions BITBUCKET_PROJECT_KEY) PASSED
bitbucket.test.sh: all tests PASSED
```

e2e regression suite (5/5 pass, both pre- and post-fix):
```
$ bash tests/host-drivers/e2e-init-bitbucket.test.sh | tail -3
  [PASS] T5: protection 403 → warn + continue; origin registered, no phase2 steps
Results: 5 passed, 0 failed
```

Self-verify gates:
```
$ bash scripts/lint-counter-antipattern.sh   # OK
$ bash scripts/lint-backlog-references.sh    # OK
```

## Bonus catches
While in `bitbucket.sh`, surfaced (but did NOT fix in this PR to keep scope tight) two adjacent S3-class issues for future remediation:

- `host_verify_protection` (lines 259-263 post-PR) captures `has_*` from `jq -r ... length` without the canonical `case … *[!0-9]* ;; esac` sanitizer. If a future malformed response causes `jq` to emit a non-numeric value, the subsequent `[ "$has_*" -eq 0 ]` integer-tests crash under `set -e`. Same defect class as the counter-antipattern that PR #72 backstopped (variant: jq-length rather than grep -c). Not introduced by this PR — pre-existing.
- The POST loop in `host_configure_protection` (`echo "$payload" | _bb_curl POST ... >/dev/null || { … }`) discards the curl stderr (already merged into stdout via `_bb_curl`'s `2>&1`) before the failure branch fires, so the operator never sees the API error body that caused the failure. The new SOIF_DEBUG buffering only covers the DELETE path; extending the same pattern to POSTs would improve diagnostic surface symmetrically. Out of scope for this PR.

## Worktree note
```
$ pwd
/Users/karl/Documents/Claude Projects/solo-orchestrator/.claude/worktrees/wf_dcd1c66b-4ea-3
```

### DIFF
--- a/docs/builders-guide.md
+++ b/docs/builders-guide.md
@@ -894,12 +894,20 @@ glab repo create <name> --private
 # init.sh handles: git setup + glab api POST projects/<path>/protected_branches
 ```
 
-**Bitbucket Cloud (first-class)** — requires an App Password (not account password) at https://bitbucket.org/account/settings/app-passwords/ with scopes `repository:admin`, `project:admin`, `pullrequest:write`, exported as env vars:
+**Bitbucket Cloud (first-class)** — requires an Atlassian API Token (App Passwords are sunset 2026; per https://support.atlassian.com/bitbucket-cloud/docs/using-api-tokens/, API tokens use HTTP Basic with the Atlassian account **email** as the username):
+```bash
+# PREFERRED — API token (create at https://id.atlassian.com/manage-profile/security/api-tokens)
+export BITBUCKET_API_TOKEN_EMAIL="you@example.com"    # Atlassian account email
+export BITBUCKET_API_TOKEN="your-api-token"
+export BITBUCKET_WORKSPACE="your-workspace-slug"
+# init.sh handles: repo create + branch-restrictions via curl
+```
+
+Legacy App Password path (sunset 2026 — still works today, will break on enforcement; scopes `repository:admin`, `project:admin`, `pullrequest:write` at https://bitbucket.org/account/settings/app-passwords/):
 ```bash
 export BITBUCKET_USER="your-bitbucket-username"
 export BITBUCKET_APP_PASSWORD="your-app-password"
-# (Org workspace? also: export BITBUCKET_WORKSPACE="org-name")
-# init.sh handles: repo create + branch-restrictions via curl
+export BITBUCKET_WORKSPACE="your-workspace-slug"
 ```
 
 **Other hosts (Gitea, Codeberg, self-hosted)**:
--- a/docs/cli-setup-addendum.md
+++ b/docs/cli-setup-addendum.md
@@ -660,22 +660,43 @@ glab auth login
 glab auth login --hostname gitlab.your-company.com
 ```
 
-### Bitbucket (curl + App Password)
+### Bitbucket (curl + API Token)
 
-No first-party CLI. Uses `curl` plus a Bitbucket App Password.
+No first-party CLI. Uses `curl` plus an Atlassian API Token.
+
+> **Why API Token, not App Password?** Atlassian is sunsetting Bitbucket
+> Cloud App Passwords in 2026 (see
+> <https://support.atlassian.com/bitbucket-cloud/docs/using-api-tokens/>).
+> API tokens are the forward-compatible replacement. Both are sent as
+> HTTP Basic authentication — for API tokens, the username is your
+> Atlassian account **email** (NOT your Bitbucket username), and the
+> password is the token itself. Bearer auth does NOT work; it is
+> reserved for OAuth 2.0 access tokens (PR #90 verifier fix).
+
+1. Create an API token at <https://id.atlassian.com/manage-profile/security/api-tokens>
+   - Scope it to Bitbucket; pick an expiry up to 1 year.
+2. Export credentials (all three are required — audit code-host-bitbucket-1):
+   ```bash
+   export BITBUCKET_API_TOKEN_EMAIL="you@example.com"    # Atlassian account email
+   export BITBUCKET_API_TOKEN="your-api-token"
+   export BITBUCKET_WORKSPACE="your-workspace-slug"
+   ```
+   `BITBUCKET_WORKSPACE` is the slug in your `bitbucket.org/<workspace>/` URL;
+   for org accounts it is the team slug, which differs from any single user.
+3. Add to your shell rc (`.bashrc` / `.zshrc`) for persistence. Ensure mode 600 if any secrets live in it.
+
+**Legacy — App Password (sunset 2026):** still works today, will break
+on Atlassian's enforcement date. Prefer the API token path above.
 
 1. Generate an App Password at <https://bitbucket.org/account/settings/app-passwords/>
    - Required scopes: `repository:admin`, `project:admin`, `pullrequest:write`
-2. Export credentials (all three are required — audit code-host-bitbucket-1):
+2. Export the three legacy vars:
    ```bash
    export BITBUCKET_USER="your-bitbucket-username"
    export BITBUCKET_APP_PASSWORD="your-app-password"
    export BITBUCKET_WORKSPACE="your-workspace-slug"
    ```
-   `BITBUCKET_WORKSPACE` is the slug in your `bitbucket.org/<workspace>/` URL.
-   For personal accounts it often (but not always) equals `BITBUCKET_USER`;
-   for org accounts it is the team slug, which differs from any single user.
-3. Add to your shell rc (`.bashrc` / `.zshrc`) for persistence. Ensure mode 600 if any secrets live in it.
+   On personal accounts `BITBUCKET_USER` often (but not always) equals the workspace slug.
 
 ### Other hosts (Gitea, Codeberg, self-hosted)
 
--- a/scripts/host-drivers/bitbucket.sh
+++ b/scripts/host-drivers/bitbucket.sh
@@ -1,16 +1,53 @@
 #!/usr/bin/env bash
 # scripts/host-drivers/bitbucket.sh — Bitbucket driver.
 # Uses curl + Bitbucket Cloud REST API 2.0.
-# Credentials via env: BITBUCKET_USER + BITBUCKET_APP_PASSWORD (App Password with
-# repository:admin, project:admin, and pullrequest:write scopes).
+#
+# Credentials (one of, in precedence order; audit code-host-bitbucket-4):
+#   1. BITBUCKET_API_TOKEN — Bitbucket Cloud API token (sent as
+#      `Authorization: Bearer …`). PREFERRED. Atlassian is sunsetting
+#      App Passwords for Bitbucket Cloud in 2026; API tokens are the
+#      forward-compatible replacement.
+#   2. BITBUCKET_APP_PASSWORD — legacy App Password (sent as HTTP Basic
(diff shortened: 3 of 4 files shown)


kraulerson:
## Adversarial review — PR #90

**Verdict: BLOCK** — one major and one substantive issue. The implementation as merged would brick the very deprecation path it claims to fix.

### Blocker — `BITBUCKET_API_TOKEN` is sent as `Bearer`; Atlassian uses HTTP Basic
`scripts/host-drivers/bitbucket.sh:37-43`

The new `_bb_auth_args` emits `-H Authorization: Bearer $BITBUCKET_API_TOKEN`. But per Atlassian's official Bitbucket Cloud API-token docs (https://support.atlassian.com/bitbucket-cloud/docs/using-api-tokens/), API tokens MUST be sent as HTTP Basic with the **Atlassian account email as the username and the token as the password** — either `curl --user 'email:token'` or `Authorization: Basic base64(email:token)`. Bearer is reserved for OAuth 2.0 access tokens, which the PR does not implement either.

Net effect: the moment Atlassian disables App Passwords (the exact event this PR claims to prepare for), every Bitbucket call returns 401. The fix produces a forward-compat illusion that fails on day one of enforcement. The unit test (`tests/host-drivers/bitbucket.test.sh:240`) asserts the wrong shape and locks in the bug, so RED-GREEN-checked all the way through.

Suggested fix: add a `BITBUCKET_API_TOKEN_EMAIL` (or reuse `BITBUCKET_USER` as the email) and emit `-u "$email:$BITBUCKET_API_TOKEN"`. Update host_require_cli to require the email alongside the token. Re-author T3 to assert `-u somebody@example.com:api-token-abc` and confirm it goes RED on the current code, then GREEN after the fix.

### Major — Promised "OR a downstream POST later fails" path is unimplemented
`scripts/host-drivers/bitbucket.sh:194-197` vs `225-232`

The new comment block promises: "buffer the diagnostic; when SOIF_DEBUG is set **OR a downstream POST later fails**, emit the buffered diagnostic so the operator sees which leftover restriction blocked creation." But the only flush site is the `SOIF_DEBUG`-gated branch at line 221-223. The POST-failure branch (`echo "bitbucket driver: failed to set $kind restriction" >&2; return 2` at 229-231) does **not** flush `$delete_diag`.

This is precisely the central use case the audit (code-host-bitbucket-3) calls out: operator runs *without* SOIF_DEBUG, a DELETE silently fails leaving a leftover, the next POST fails with "failed to set force restriction", and the operator still has no diagnostic about the cleanup failure that caused it. Fix closes the partial case (operator already runs with SOIF_DEBUG) but misses the common case (operator hits the bug, *then* re-runs with debug, having lost the original timing). T2 only exercises the SOIF_DEBUG path, so the missing branch is invisible to the test suite.

Suggested fix: in the POST failure `{}` block, prepend `[ -n "$delete_diag" ] && printf '%s' "$delete_diag" >&2` before the existing error message. Add a T2b that omits SOIF_DEBUG, fails one DELETE *and* one POST, and asserts the delete-diagnostic surfaces.

### Minor — User-facing setup docs still document only App Passwords
`docs/cli-setup-addendum.md:671-676`, `docs/builders-guide.md:899-901`

Runtime `host_require_cli` was updated, but the static setup docs were not. A new Bitbucket adopter following the published setup guide still configures the deprecated App-Password path. Update both docs to lead with `BITBUCKET_API_TOKEN` and demote App Password to a "legacy / sunset 2026" note.

### Minor — Comment promise mismatch
`scripts/host-drivers/bitbucket.sh:29-36` says auth tokens are "space-separated" but the function actually emits them newline-separated (which is what the caller's `while IFS= read -r tok` requires). Docstring should say "one per line."

### Nit — `_bb_auth_args` silently emits nothing when both env vars are unset
`scripts/host-drivers/bitbucket.sh:37-43` — if both BITBUCKET_API_TOKEN and BITBUCKET_APP_PASSWORD are empty (host_require_cli is the only gate), the function emits no args and `curl -sSf` proceeds unauthenticated. Belt-and-braces would be to early-fail in `_bb_curl` / `_bb_curl_no_body` when `auth=()`.

### Verified clean
- Test suite: full host-driver `run-all.sh` is 9/9 green on PR branch (re-ran in clean worktree); no hidden regressions.
- TDD honesty: T1 (`could not list existing restrictions`) does fail on `main` (`ASSERT FAIL: 'bitbucket driver: failed to set force restriction' does not contain 'could not list existing restrictions'`). Confirmed RED→GREEN.
- `lint-counter-antipattern.sh` passes.
- Pre-existing `host_verify_protection` jq-length sanitizer gap (lines 259-263) correctly flagged as bonus catch and out of scope.

kraulerson:
## Adversarial verifier BLOCK addressed — fixup commit e4c9777

All five issues from the BLOCK verdict are fixed in commit
`e4c9777a7d0da1e16c7d363f18f6c1938b31e22f`.

### BLOCKER (1) — API token now uses HTTP Basic, not Bearer
Validated via Context7 against Atlassian's official
[Bitbucket Cloud API tokens guide](https://support.atlassian.com/bitbucket-cloud/docs/using-api-tokens/):
API tokens are sent as HTTP Basic with the Atlassian account
**email** as the username and the token as the password. Bearer is
reserved for OAuth 2.0 access tokens.

- Added `BITBUCKET_API_TOKEN_EMAIL` (kept distinct from
  `BITBUCKET_USER`, which is the Bitbucket username used for legacy
  App Password Basic auth — the two are not always the same).
- `_bb_auth_args` now emits `-u $BITBUCKET_API_TOKEN_EMAIL:$BITBUCKET_API_TOKEN`.
- `host_require_cli` requires both halves of whichever pair is used;
  guidance text leads with the API-token + email pair.
- Re-authored T3 to assert the `-u email:token` shape; added T5b for
  the missing-email failure mode.

TDD red→green for the auth fix (commit body has the full traces):

  RED  — ASSERT FAIL [API token uses HTTP Basic with email as user]:
         '… -H Authorization: Bearer api-token-abc …' does not contain
         '-u user@example.com:api-token-abc'
  GREEN — bitbucket.test.sh: API token uses HTTP Basic with email (no Bearer) PASSED

### MAJOR (2) — POST-failure path now flushes buffered delete diagnostic
`host_configure_protection`'s POST-failure block now prepends
`[ -n "$delete_diag" ] && printf '%s' "$delete_diag" >&2` before the
existing "failed to set <kind> restriction" message. Added T2b
(DELETE fails AND POST fails, `SOIF_DEBUG` unset, asserts the
delete diagnostic surfaces).

  RED  — ASSERT FAIL [delete diagnostic flushed on POST failure
         (no SOIF_DEBUG)]: '… failed to set force restriction' does
         not contain 'restriction 77'
  GREEN — bitbucket.test.sh: host_configure_protection (POST failure
         flushes delete diag without SOIF_DEBUG) PASSED

### MINOR (3) — Setup docs lead with API Token; App Password = legacy
- `docs/cli-setup-addendum.md` rewritten to lead with the API-token
  pair, link to Atlassian's guide, and demote App Password to a
  "Legacy — sunset 2026" subsection.
- `docs/builders-guide.md` updated with the same restructure.

Note: Atlassian has not yet published a single calendar date beyond
"2026" for App Password sunset, so both docs keep the existing
"sunset 2026" phrasing rather than invent a more specific date.

### MINOR (4) — `_bb_auth_args` docstring corrected
"space-separated" → "one per line" + clarified both paths are HTTP
Basic now.

### NIT (5) — Empty-auth early-fail in `_bb_curl` / `_bb_curl_no_body`
Both wrappers now check `${#auth[@]} -eq 0` and refuse to issue a
request with a clear diagnostic if `host_require_cli` was somehow
bypassed. (Inlined per call site to stay bash-3.2 compatible —
`local -n` is bash 4+.)

### Self-verify
- `bash tests/host-drivers/bitbucket.test.sh` — 22/22 PASSED
- `bash tests/host-drivers/run-all.sh` — 9/9 suites green
- `bash scripts/lint-counter-antipattern.sh` — OK
- `bash scripts/lint-fixture-envelopes.sh` — OK
- `bash scripts/lint-backlog-references.sh` — OK

Diff stat: 4 files changed, 170 insertions(+), 59 deletions(-).

PR #92, opened 2026-06-28
fix(verify-install): fix_claude_md stub + check_hooks on-disk + fix_tool_install eval safety (3 S3 closures)

kraulerson:
## Summary
- `scripts/verify-install.sh` had three independent S3 defects sharing a root cause: fix/check functions reported success without verifying the side-effects they claimed. Each closure also exposed a sibling silent-degradation pattern that the PR catches as bonus fixes.
- `fix_claude_md` now uses `templates/generated/claude-md.tmpl` + the same sed substitution recipe init.sh:2230-2247 uses, appending the organizational Branch Protection block. Refuse-on-missing-template + atomic mktemp+mv write — no half-CLAUDE.md on SIGINT.
- `check_hooks` was JSON-only; now requires settings.json reference **AND** on-disk script (`[ -e ]` + `[ -x ]`) before emitting `register_pass`. Applied to all 8 hook-script rows (pre-commit-gate, track-tool-usage, two SessionStart scripts, session-end-qdrant-reminder, bypass-detector × 2 events, record-claude-commit, detect-out-of-band-commits).
- `fix_tool_install` no longer passes resolver-supplied strings to bare `eval ... 2>/dev/null`. Defense in depth: (a) allowlist the first argv token against the 18 package-manager invocation shapes already shipped in `templates/tool-matrix/*.json`; (b) echo the resolved command to stderr before executing; (c) drop the `2>/dev/null` mask so install-time failures are visible. Sibling cleanup in `run_remediation` (`$fix_func 2>/dev/null` → `$fix_func`) so the audit trail actually reaches the operator.
- 14-scenario regression test (`tests/test-verify-install-fix-functions.sh`) covers all three findings + bonus catches. Existing `tests/test-verify-install-bl030-coverage.sh` still green (8/8).

## Findings closed
- `code-verify-reconfigure-9` — `fix_claude_md` regenerated CLAUDE.md is now at template parity with init.sh (pending-approval sentinel bullet, deployment context, Build-Loop / persona / CDF guidance, organizational Branch Protection appendix).
- `code-verify-reconfigure-11` — `check_hooks` verifies on-disk presence + execute bit for all hook scripts, not just settings.json references.
- `code-verify-reconfigure-14` — `fix_tool_install` allowlists the first argv token, echoes the resolved command, and surfaces install-time stderr instead of masking it. Sibling stderr-silencer in `run_remediation` also removed.

## Test plan

Red against worktree base (pre-fix):

```
$ bash tests/test-verify-install-fix-functions.sh
T1: fix_claude_md regenerates CLAUDE.md with pending-approval-sentinel bullet
  [FAIL] T1 — regenerated CLAUDE.md missing the pending-approval-sentinel bullet
...
T4: organizational deployment appends Branch Protection block
  [FAIL] T4 — organizational regenerated CLAUDE.md missing Branch Protection block
...
T6-T10: missing on-disk hook scripts — all reported [FAIL] (PASS row leaked through)
...
Results: 5 passed, 8 failed
```

Green after fix:

```
$ bash tests/test-verify-install-fix-functions.sh
T1  [PASS] regenerated CLAUDE.md contains pending-approval-sentinel bullet
T2  [PASS] identity fields substituted (project/platform/track/language)
T3  [PASS] no unsubstituted __TOKEN__ placeholders remain
T4  [PASS] organizational Branch Protection block appended
T5  [PASS] no token leak when source template unavailable
T6  [PASS] hook with missing on-disk script no longer reported as PASS
T7  [PASS] hook with non-executable on-disk script no longer reported as PASS
T8  [PASS] PostToolUse track-tool-usage with missing on-disk script no longer PASS
T9  [PASS] SessionStart hooks no longer PASS when on-disk script is missing
T10 [PASS] Stop hook session-end-qdrant-reminder no longer PASS when on-disk script is missing
T11 [PASS] fix_tool_install did NOT execute attacker-controlled install_cmd
T12 [PASS] refusal reason emitted to stderr
T13 [PASS] allowlisted brew install shape is accepted (shim invoked)
T14 [PASS] command echoed to stderr before execution
Results: 14 passed, 0 failed
```

Regression coverage:

```
$ bash tests/test-verify-install-bl030-coverage.sh
T1-T8: all [PASS]
Results: 8 passed, 0 failed
```

CI lints:

```
```

## Bonus catches
- `run_remediation` silenced stderr from every fix function via `if $fix_func 2>/dev/null; then` (`scripts/verify-install.sh:1133`). This undermined the audit-trail my `fix_tool_install` cleanup added. Removed the silencer; fix functions that need to suppress noisy stderr should do so locally.
- `load_context` skipped `.claude/phase-state.json` when determining `DEPLOYMENT`. The pre-existing fallback was `.claude/intake-progress.json` → `CLAUDE.md` grep — but in the exact case `fix_claude_md` is invoked to repair (CLAUDE.md missing/deleted), the grep returned nothing and `DEPLOYMENT` stayed empty. Result: the organizational Branch Protection appendix was silently skipped. Added phase-state.json as an intermediate trust source (init.sh always writes `.deployment` there).
- Three other fix functions (`fix_framework_clone`, `fix_framework_manifest`, `fix_superpowers`) carry the same `2>/dev/null` silencer pattern but weren't in scope for this PR — flagged for follow-up.

## Worktree note
```
$ pwd
/Users/karl/Documents/Claude Projects/solo-orchestrator/.claude/worktrees/wf_dcd1c66b-4ea-1
```

### DIFF
--- a/init.sh
+++ b/init.sh
@@ -1230,6 +1230,14 @@ create_project() {
   cp "$SCRIPT_DIR/templates/generated/changelog.tmpl" templates/generated/
   cp "$SCRIPT_DIR/templates/generated/bugs.tmpl" templates/generated/
   cp "$SCRIPT_DIR/templates/generated/release-notes.tmpl" templates/generated/
+  # Verifier nit-2 follow-up (PR #92 review): copy the canonical
+  # claude-md.tmpl into the project so verify-install.sh's self-
+  # bootstrap fallback in fix_claude_md (scripts/verify-install.sh:645
+  # — `elif [ -f "templates/generated/claude-md.tmpl" ]`) has something
+  # to read when $SOURCE_DIR is no longer reachable (developer rotated
+  # the orchestrator clone, CI restored only the project tree, etc.).
+  # Without this copy that fallback was dead code in practice.
+  cp "$SCRIPT_DIR/templates/generated/claude-md.tmpl" templates/generated/
 
   # Install vendored skills (project-level, .claude/skills/<name>/).
   # Skills are markdown SKILL.md files with NOTICE-attribution preserved.
--- a/scripts/verify-install.sh
+++ b/scripts/verify-install.sh
@@ -337,6 +337,53 @@ check_git() {
   fi
 }
 
+# Audit code-verify-reconfigure-11 helper. Moved from inside check_hooks
+# to file scope (verifier nit-1 follow-up): bash function definitions
+# inside another function are hoisted to global scope on the first
+# invocation of the parent, so the prior placement only saved style
+# points and re-paid the function-registration cost per check_hooks
+# call. File-scope placement matches the convention used by sibling
+# helpers (run_with_timeout, _tool_install_head_allowed, etc.) and
+# documents the helper as part of verify-install's hook contract
+# rather than as a local detail of check_hooks.
+#
+# _check_hooks_hook_pair: validates one (event-jq-match, on-disk-path)
+# pair and emits the appropriate register_pass / register_manual row.
+# Arguments:
+#   $1 = human label (e.g. "PreToolUse hook: pre-commit-gate.sh")
+#   $2 = jq -e expression matching the hook entry in settings.json
+#   $3 = on-disk script path (relative to project root)
+#   $4 = manual remediation instruction string
+_check_hooks_hook_pair() {
+  local label="$1"
+  local jq_match="$2"
+  local script_path="$3"
+  local remediation="$4"
+
+  if ! jq -e "$jq_match" .claude/settings.json >/dev/null 2>&1; then
+    register_manual "$label not registered" "$remediation"
+    return
+  fi
+
+  # Registration present — now validate the on-disk script. Reuse the
+  # check_scripts row when one is already emitted for $script_path
+  # (e.g. pre-commit-gate.sh is in the canonical scripts array), but
+  # we still want a hook-row signal so operators can correlate the
+  # JSON ref with the runtime requirement.
+  if [ ! -e "$script_path" ]; then
+    register_manual "$label registered but on-disk script missing ($script_path)" \
+      "Restore $script_path from the orchestrator source, then re-run verify-install"
+    return
+  fi
+  if [ ! -x "$script_path" ]; then
+    register_manual "$label registered but on-disk script not executable ($script_path)" \
+      "Run: chmod +x $script_path"
+    return
+  fi
+
+  register_pass "$label"
+}
+
 check_hooks() {
   print_step "Checking Claude Code hook registration..."
 
@@ -359,50 +406,14 @@ check_hooks() {
   # a Claude Code PreToolUse hook; on-disk presence is the load-bearing
   # detail. The fix below requires BOTH the JSON registration AND the
   # on-disk script to be present + executable before a PASS is emitted.
-  #
-  # _check_hook_pair: validates one (event-jq-match, on-disk-path) pair
-  # and emits the appropriate register_pass / register_manual row.
-  # Arguments:
-  #   $1 = human label (e.g. "PreToolUse hook: pre-commit-gate.sh")
-  #   $2 = jq -e expression matching the hook entry in settings.json
-  #   $3 = on-disk script path (relative to project root)
-  #   $4 = manual remediation instruction string
-  _check_hook_pair() {
-    local label="$1"
-    local jq_match="$2"
-    local script_path="$3"
-    local remediation="$4"
-
-    if ! jq -e "$jq_match" .claude/settings.json >/dev/null 2>&1; then
-      register_manual "$label not registered" "$remediation"
-      return
-    fi
-
-    # Registration present — now validate the on-disk script. Reuse the
-    # check_scripts row when one is already emitted for $script_path
-    # (e.g. pre-commit-gate.sh is in the canonical scripts array), but
-    # we still want a hook-row signal so operators can correlate the
-    # JSON ref with the runtime requirement.
-    if [ ! -e "$script_path" ]; then
-      register_manual "$label registered but on-disk script missing ($script_path)" \
-        "Restore $script_path from the orchestrator source, then re-run verify-install"
-      return
-    fi
-    if [ ! -x "$script_path" ]; then
-      register_manual "$label registered but on-disk script not executable ($script_path)" \
-        "Run: chmod +x $script_path"
-      return
-    fi
-
-    register_pass "$label"
-  }
+  # The per-row helper lives at file scope as _check_hooks_hook_pair.
 
   # PreToolUse hook: pre-commit-gate.sh — also verify matcher is Bash.
   if jq -e '.hooks.PreToolUse[]? | .hooks[]? | select(.command | contains("pre-commit-gate.sh"))' .claude/settings.json >/dev/null 2>&1; then
     local matcher
     matcher=$(jq -r '[.hooks.PreToolUse[]? | select(.hooks[]? | .command | contains("pre-commit-gate.sh")) | .matcher // "none"] | first' .claude/settings.json 2>/dev/null || echo "unknown")
     if [ "$matcher" = "Bash" ]; then
-      _check_hook_pair \
+      _check_hooks_hook_pair \
         "PreToolUse hook: pre-commit-gate.sh (matcher: Bash)" \
         '.hooks.PreToolUse[]? | .hooks[]? | select(.command | contains("pre-commit-gate.sh"))' \
         "scripts/pre-commit-gate.sh" \
@@ -417,7 +428,7 @@ check_hooks() {
   fi
 
   # PostToolUse hook: track-tool-usage.sh
-  _check_hook_pair \
+  _check_hooks_hook_pair \
     "PostToolUse hook: track-tool-usage.sh" \
     '.hooks.PostToolUse[]? | .hooks[]? | select(.command | contains("track-tool-usage.sh"))' \
     "scripts/track-tool-usage.sh" \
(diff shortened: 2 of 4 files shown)


kraulerson:
## Adversarial review — PR #92

Verdict: **major_concerns**. The fix_claude_md and check_hooks closures are sound and well-implemented. The fix_tool_install closure is materially weaker than the audit recommended, and two of the new tests do not actually red-state the bug they claim to cover. Findings below.

### Blocker-class

**1. The first-token allowlist is trivially bypassable via shell-metacharacter chaining — finding code-verify-reconfigure-14 is only partially closed.** (`scripts/verify-install.sh:989-1052`)

The allowlist gates only the leading argv token, then the whole resolver-supplied string is handed to `bash -c -- \"$install_cmd\"`. Any allowlisted head (`brew`, `npm`, `sudo`, `curl`, `source`, `echo`, ...) followed by `;`, `&&`, `||`, `|`, `` ` ``, or `$(...)` chains arbitrary commands. Repro on the PR branch:

```
$ RESOLVER_OUTPUT='{\"auto_install\":[{\"install_cmd\":\"brew --version; touch /tmp/X\"}]}'
$ fix_tool_install 0
[INFO] fix_tool_install: executing brew --version; touch /tmp/X
Homebrew 6.0.5
$ ls /tmp/X
/tmp/X   # injection succeeded
```

`sudo` is in the allowlist — `sudo any-arbitrary-command` is one supply-chain entry away. The PR body candidly acknowledges this is a \"defense-in-depth concession\" and the structured `package_manager` dispatch is deferred. That's a reasonable engineering call, but the audit finding (\"unbounded eval on JSON-supplied install commands → RCE\") is mitigated only to \"RCE if the attacker leads with an allowlisted token,\" which is the path of least resistance. The closure should either be downgraded (\"partial fix; structured dispatch tracked separately\") or T-test should explicitly cover the chained-injection bypass to surface the residual risk.

The new tests T11/T13 do not exercise this — T11 uses a non-allowlisted head (`touch`), T13 uses a clean `brew install ...`. Suggest adding T11b: payload `\"brew --version; touch \$MARKER\"` asserting the marker IS created (documenting the known weakness) or refusal (forcing the structured dispatch now).

### Major

**2. T11 passes vacuously on `origin/main` — does not red-state the bug it claims to cover.** (`tests/test-verify-install-fix-functions.sh:691-706`)

The `_invoke_fix_tool_install` awk extractor looks for `^_TOOL_INSTALL_ALLOWED_HEADS=\\(` — that symbol does not exist on `main`. Running the extractor against `origin/main:scripts/verify-install.sh`:

```
$ awk '/^_TOOL_INSTALL_ALLOWED_HEADS=\\(/ {flag=1} flag {print} ...' /tmp/main.sh | wc -l
0
```

So on main: extract is empty, `fix_tool_install` is undefined, never invoked, marker never created → **T11 PASSES on main without touching the vulnerable `eval` code path at all**. The PR body claims \"T11 fails on the pre-fix base\" — that's only true because the test never reaches main's `eval`. T12/T13/T14 do legitimately fail on main (no allowlist output, no shim invocation). To actually red-state T11, the extractor needs a fallback that loads main's `fix_tool_install` (e.g. detect missing allowlist → source the function body directly via a second awk pass).

**3. T5 is vacuous — passes on main and the PR branch alike.** (`tests/test-verify-install-fix-functions.sh:562-582`)

T5 asserts \"no `__[A-Z_]*__` tokens leaked when source template unavailable.\" On main, `fix_claude_md` uses a bash heredoc with `${PROJECT_NAME:-unknown}` interpolation — there are no `__TOKEN__` strings in main's stub either. So T5 passes on main with NO source-template path involved. The test does not exercise the new branch (`tmpl=\"\"` → return 1, no CLAUDE.md written). Suggest replacing with: assert `[ ! -f \"\$PROJ/CLAUDE.md\" ]` AND assert a manual-action message was registered. That would actually distinguish the new safe-refusal behavior from main's placebo-stub behavior.

### Minor

**4. Bonus catch is incomplete — three sibling fix functions still carry `$func 2>/dev/null` silencers.** (`scripts/verify-install.sh:947-1067`)

The PR removed the dispatcher-level `$fix_func 2>/dev/null` in `run_remediation` (good), but `fix_framework_clone` (line 947), `fix_framework_manifest` (line 952), and `fix_superpowers` (line 1067) still suppress stderr internally per the PR body's own \"flagged for follow-up\" note. Flagging follow-up is fine; a tracked backlog row / open issue would be better than a buried mention in the PR description.

**5. T6/T7 specifically test pre-commit-gate.sh, but the test never re-stages the on-disk file between scenarios.** Not a bug here because `teardown` resets to a fresh project, but if the test is reorganized to share a single fixture, the chmod/rm side-effects will cascade.

### Nit

**6. `_check_hook_pair` is defined inside `check_hooks()`** (line 371). Bash supports nested function definitions, but the function becomes globally visible once `check_hooks` runs. Moving it to file-scope alongside `check_hooks` (and naming it `_check_hooks_hook_pair`) would match the project's existing helper conventions and avoid a re-definition cost per invocation.

**7. `templates/generated/claude-md.tmpl` is not copied to the project during `init.sh` (lines 1220-1232 cp other templates but skip claude-md.tmpl).** The fallback `elif [ -f \"templates/generated/claude-md.tmpl\" ]` (line 645) is therefore dead code in practice. Harmless, but worth a comment or a follow-up to either remove the fallback or have init.sh also copy the canonical template into the project tree.

### What is solid

- code-verify-reconfigure-9 fix is faithful to init.sh's recipe; atomic mktemp+mv is correct; the load_context bonus (phase-state.json fallback for DEPLOYMENT) is a legitimate latent bug fix.
- code-verify-reconfigure-11 fix is comprehensive — eight hook rows now require JSON + on-disk + +x. T6-T10 are real red→green tests, and the SessionStart paired-script handling preserves the existing semantics.
- The run_remediation stderr unmask is a clean improvement.
- Lint counter-antipattern and lint-backlog-references both still pass on the branch.

kraulerson:
## Verifier-driven fix commit pushed (`6917fd2`)

All six adversarial-review findings addressed.

### BLOCKER — first-token allowlist bypassable via metachar chaining

`scripts/verify-install.sh::fix_tool_install` rewritten as **two-layer dispatch**:

- **Layer 1 (preferred — STRUCTURED):** recognizes `<pkg-mgr> install <pkg>` shapes for brew, sudo apt|dnf, sudo pacman -S, npm i/-g, pip|pip3|pipx, cargo, gem. Dispatches via direct argv (`brew install -- "$pkg"`) — **no `bash -c`, no `eval`**. Package token validated against a strict regex. The verifier's repro (`brew --version; touch /tmp/X`) does NOT match a structured shape and falls through to Layer 2.

- **Layer 2 (deprecated — legacy string path):** kept behind `VERIFY_INSTALL_NO_LEGACY_DISPATCH=1` opt-out, emits DEPRECATED warning before each fall-through. New **metachar gate** REFUSES any post-allowlist payload containing `;`, `|`, `` ` ``, `$(`, `<`, `>`, newline, or bare `&`. `&&` and `||` are stripped before the check so legitimate multi-stage installs (e.g. `brew install foo && brew services start bar`) still work. The verifier's repro now lands here and is refused by the metachar gate with a visible reason.

- **T11b added** — asserts `brew --version; touch $MARKER` is REFUSED. Confirmed RED on origin/main's `eval`-based path; GREEN after the fix.

### MAJOR (1) — T11 vacuous on `origin/main`

`_invoke_fix_tool_install` extractor now has a **two-pass strategy**: pass 1 looks for the post-fix `_TOOL_INSTALL_ALLOWED_HEADS=(` anchor; pass 2 falls back to the bare `fix_tool_install() {` body that lives on `main`. Confirmed RED on main: extracted main `fix_tool_install` + T11 payload creates the marker via main's `eval`.

### MAJOR (2) — T5 vacuous

T5 oracle replaced with the verifier's suggested signal: assert `[ ! -f $PROJ/CLAUDE.md ]` AND assert a CLAUDE.md remediation row was registered. Main's heredoc stub ALWAYS writes CLAUDE.md, so the new T5 RED-states on main and GREEN-states on the safe-refusal path. Also removes the in-project template copy as part of test setup so the self-bootstrap fallback misses too.

### MINOR — incomplete `2>/dev/null` scrub

Dropped `2>/dev/null` from `fix_framework_clone`, `fix_framework_manifest`, `fix_superpowers`. Each carries a comment cross-referencing the code-verify-reconfigure-14 rationale.

### NIT (1) — `_check_hook_pair` nested inside `check_hooks`

Moved to file scope as `_check_hooks_hook_pair`. All seven callsites updated.

### NIT (2) — `claude-md.tmpl` never copied during `init.sh`

`init.sh` now copies `claude-md.tmpl` into the project's `templates/generated/` alongside the other documentation templates. The `verify-install.sh:645` fallback is no longer dead code.

### Backlog

New **BL-033** tracking migration of multi-stage `install_cmd` entries in `templates/tool-matrix/*.json` that now trip the metachar gate (gitleaks, k6, rust, context7). Workaround: operators see DEPRECATED + refusal and use the matrix's `manual` instructions until migrated.

### RED→GREEN evidence

**RED on `origin/main`** (T5, T11, T11b run against main's `verify-install.sh` + `init.sh` via swap-and-restore):
```
T5:   [FAIL] CLAUDE.md was written despite no canonical template available
T11:  [FAIL] marker created
T11b: [FAIL] chained-injection executed (marker created)
Results: 0 passed, 3 failed
```

**GREEN on PR branch** (`tests/test-verify-install-fix-functions.sh`, all 15 tests):
```
T1-T4   PASS (fix_claude_md template parity)
T5      PASS (safe-refusal — CLAUDE.md absent and remediation row registered)
T6-T10  PASS (check_hooks on-disk + +x)
T11     PASS (non-allowlisted refused)
T11b    PASS (chained-injection REFUSED with visible reason)
T12-T14 PASS (refusal-reason / structured-dispatch / audit-trail)
Results: 15 passed, 0 failed
```

**Lints**: `scripts/lint-counter-antipattern.sh` OK · `scripts/lint-backlog-references.sh` OK.

Stats: 4 files changed, 420 insertions(+), 115 deletions(-).

PR #93, opened 2026-06-29
fix(lib,hooks): host.sh date parse + bypass-audit atomicity + bypass-detector session_id (3 S3 closures)

kraulerson:
## Summary

- **code-lib-1**: `host_verify_protection` on the `other`-host fallback chained `date -d` then `date -j -f` then `|| echo "$now"`. The terminal fallback silently classified any unparseable attestation timestamp as fresh (`days=0`), bypassing the 90-day W3 staleness check. Replaced with an explicit dual-parser-failure branch that emits a named-value stderr warning and `return 1` (fail-closed).
- **code-lib-2**: `bypass_audit_append` + `bypass_audit_close_pending` were already using the adjacent-mktemp atomic-rename pattern (PR #46/0d5605e). Added the two residual hardenings from the original recommendation: (a) preserve target file permissions across the rename so operator-set perms (e.g. 0640) don't silently revert to mktemp's 0600 default — portable `_bypass_audit_preserve_mode` helper (GNU `chmod --reference` → BSD/GNU `stat`-then-`chmod` → `chmod 600` fallback); (b) EXIT/INT/TERM trap that removes the tmp file + lockdir if a signal hits between mktemp and the rename, so the governance dir doesn't accumulate orphan `${file}.XXXXXX` litter.
- **code-hooks-bypass-detector-3**: `bypass-detector.sh` line 107 read `${CLAUDE_SESSION_ID:-unknown}` — but Claude Code does NOT export that env var. The documented correlator is the envelope's top-level `.session_id` field. Every audit row was landing with `session_id="unknown"`, breaking W7 successor-handoff correlation. Now reads `$(echo "$INPUT" | jq -r '.session_id // empty')`, falling back to `CLAUDE_SESSION_ID` (belt-and-suspenders for unexpected envelope shapes), then `"unknown"`.

## Findings closed

- code-lib-1 — host.sh::host_verify_protection unparseable timestamp silently fresh
- code-lib-2 — bypass-audit tmp-file hardening (perm-preserve + EXIT trap)
- code-hooks-bypass-detector-3 — session_id read from undocumented env var

### RED (against worktree base, before fixes)

```
$ bash tests/test-host-verify-protection-date-parse.sh
T1: fresh ISO-8601 attestation verifies (returns 0)
  [PASS] T1
T2: 180-day-old attestation fails verification (returns 1)
  [PASS] T2
T3: unparseable timestamp returns 1 (fail-closed, not silently fresh)
  [FAIL] T3 — expected non-zero (fail-closed); got rc=0 — silent bypass restored
T4: stderr warning names the unparseable value
  [FAIL] T4 — expected stderr to name 'unparseable' + the value; got: 
T5: ISO-8601 with explicit 'Z' parses cleanly on both GNU and BSD
  [PASS] T5
Results: 3 passed, 2 failed

$ bash tests/test-bypass-audit-tmp-hardening.sh
T1: 0640 audit file keeps 0640 after append
  [FAIL] T1 — expected 640 after append; got 600 (mktemp 0600 leaked through rename)
T2: 0640 audit file keeps 0640 after close_pending
  [FAIL] T2 — expected 640 after close_pending; got 600
T3: append on a fresh init produces mode-600 file (not umask default)
  [FAIL] T3 — mode churn: pre=644 post=600
T4: bypass_audit_append registers an EXIT trap for tmp cleanup
  [FAIL] T4 — bypass_audit_append has no EXIT/INT/TERM trap protecting $tmp
T5: bypass_audit_close_pending registers an EXIT trap for tmp cleanup
  [FAIL] T5 — bypass_audit_close_pending has no EXIT/INT/TERM trap protecting $tmp
T6: no orphan tmp files after a successful append
  [PASS] T6
T7: no orphan tmp files after a successful close_pending
  [PASS] T7
Results: 2 passed, 5 failed

$ bash tests/test-bypass-detector-session-id.sh
T1: envelope .session_id is written to the audit row
  [FAIL] T1 — expected 'sess-abc-123'; got 'unknown'
T2: envelope .session_id wins over CLAUDE_SESSION_ID env var
  [FAIL] T2 — expected 'envelope-wins'; got 'env-fallback-id'
T3: missing envelope .session_id → $CLAUDE_SESSION_ID fallback
  [PASS] T3
T4: both missing → 'unknown' sentinel
  [PASS] T4
T5: empty envelope .session_id → env var fallback
  [PASS] T5
Results: 3 passed, 2 failed
```

### GREEN (after fixes)

```
$ bash tests/test-host-verify-protection-date-parse.sh
T1 [PASS] | T2 [PASS] | T3 [PASS] | T4 [PASS] | T5 [PASS]
Results: 5 passed, 0 failed

$ bash tests/test-bypass-audit-tmp-hardening.sh
T1 [PASS] (was 640, is 640) | T2 [PASS] | T3 [PASS] (644 unchanged)
T4 [PASS] | T5 [PASS] | T6 [PASS] | T7 [PASS]
Results: 7 passed, 0 failed

$ bash tests/test-bypass-detector-session-id.sh
T1 [PASS] (envelope) | T2 [PASS] (envelope wins) | T3 [PASS] (env fallback)
T4 [PASS] (unknown) | T5 [PASS] (empty → env)
Results: 5 passed, 0 failed
```

### Regressions (all GREEN)

- `tests/test-bypass-audit-lib.sh`           — 10/10
- `tests/test-bypass-audit-integrity.sh`     — 8/8 (D1/D2/D3 sections)
- `tests/test-bypass-audit-schema.sh`        — 3/3
- `tests/test-bypass-detector.sh`            — 15/15
- `tests/test-bypass-sentinel.sh`            — 5/5
- `tests/test-bl029-integration.sh`          — 6/6
- `tests/test-init-other-host-attestation.sh` — 2/2
- `tests/test-check-phase-gate-backstop-attestation.sh` — 3/3

### Lints

```
```

## Bonus catches

- `_bypass_audit_preserve_mode` is intentionally extracted as a reusable helper. Adjacent code (`scripts/reconfigure-project.sh`, `scripts/intake-wizard.sh`, `scripts/pending-approval.sh`, `scripts/upgrade-project.sh`) has 20+ `tmp=$(mktemp)` writes that would benefit from the same chmod-preserve + trap discipline; not pulled into this PR (out of scope for the 3-finding bundle) but the helper is in place for a follow-up sweep.
- The detector's `session_id` regression suggests the broader hook-envelope-vs-env-var schema is worth a one-shot grep: any other hook reading `$CLAUDE_*_ID` env vars instead of the documented envelope field is likely similarly broken. Not in this PR's scope; flagging for the cross-cutting Wave 3 sweep slot.

## Worktree note

```
$ pwd
/Users/karl/Documents/Claude Projects/solo-orchestrator/.claude/worktrees/wf_0dedbb55-0f8-4
```

### DIFF
--- a/scripts/hooks/bypass-detector.sh
+++ b/scripts/hooks/bypass-detector.sh
@@ -104,7 +104,15 @@ PATTERNS=$(scan_bypass_patterns_all "$TEXT" || true)
 
 # Build the rows.
 TS=$(date -u +"%Y-%m-%dT%H:%M:%SZ")
-SESSION_ID="${CLAUDE_SESSION_ID:-unknown}"
+# Audit fix code-hooks-bypass-detector-3 (2026-06-28): the Claude Code
+# hook envelope carries .session_id as a documented top-level field;
+# CLAUDE_SESSION_ID is NOT exported by Claude Code, so the previous
+# env-only read landed every row with session_id="unknown" and broke
+# the W7 successor-handoff correlation. Read the envelope first; fall
+# back to $CLAUDE_SESSION_ID (belt-and-suspenders for unexpected
+# envelope shapes), then to "unknown".
+SESSION_ID=$(echo "$INPUT" | jq -r '.session_id // empty' 2>/dev/null)
+[ -z "$SESSION_ID" ] && SESSION_ID="${CLAUDE_SESSION_ID:-unknown}"
 LEVEL=$(jq -r '.enforcement_level // "strict"' "$PROJECT_ROOT/.claude/manifest.json" 2>/dev/null)
 FIRST_PATTERN=""
 
--- a/scripts/lib/bypass-audit.sh
+++ b/scripts/lib/bypass-audit.sh
@@ -44,12 +44,21 @@ _bypass_audit_preserve_mode() {
 # bypass_audit_init <project_root>
 # Creates .claude/bypass-audit.json as an empty array if it does not already
 # exist. Idempotent — preserves existing rows.
+#
+# Verifier follow-up (2026-06-28): chmod 600 after creation. The plain
+# `echo "[]" > "$file"` redirect inherits the caller's umask (commonly
+# 0022 → file mode 0644), which leaves the governance ledger world-
+# readable on a default-umask multi-user box. Forcing 0600 here gives
+# _bypass_audit_preserve_mode a sane baseline to copy from on the first
+# append, so an umask-derived leak doesn't perpetuate through later
+# rename cycles.
 bypass_audit_init() {
   local project_root="${1:-.}"
   local file="$project_root/.claude/bypass-audit.json"
   [ -f "$file" ] && return 0
   mkdir -p "$project_root/.claude"
   echo "[]" > "$file"
+  chmod 600 "$file" 2>/dev/null || true
 }
 
 # bypass_audit_append <project_root> <row_json>
@@ -102,21 +111,35 @@ bypass_audit_append() {
   #     either platform.
   # (b) Trap on EXIT/INT/TERM so a signal between mktemp and either
   #     branch doesn't leave an orphan ${file}.XXXXXX littering the
-  #     governance dir. The trap clears itself after successful
-  #     rename so the lock_dir cleanup runs unimpeded.
-  local tmp rc
-  tmp=$(mktemp "${file}.XXXXXX")
-  trap 'rm -f "$tmp"; rmdir "$lock_dir" 2>/dev/null' EXIT INT TERM
-  if jq --argjson r "$row" '. + [$r]' "$file" > "$tmp" 2>/dev/null; then
-    _bypass_audit_preserve_mode "$file" "$tmp"
-    mv "$tmp" "$file"
-    rc=0
-  else
-    rm -f "$tmp"
-    echo "[FAIL] bypass_audit_append: jq failed" >&2
-    rc=1
-  fi
-  trap - EXIT INT TERM
+  #     governance dir.
+  #
+  # Verifier follow-up (2026-06-28): wrap the rename window in a
+  # SUBSHELL. bash `trap` is shell-global, not function-local — the
+  # prior `trap '...' EXIT INT TERM` / `trap - EXIT INT TERM` pair at
+  # function scope silently destroyed any pre-existing EXIT trap the
+  # caller had installed. Containing the trap in a subshell means it
+  # only governs that subshell's exit; the caller's trap survives
+  # untouched. tmp/rc must be captured from the subshell's exit code
+  # since locals don't propagate up.
+  local rc=0
+  (
+    tmp=$(mktemp "${file}.XXXXXX") || exit 1
+    trap 'rm -f "$tmp"; rmdir "$lock_dir" 2>/dev/null' EXIT INT TERM
+    if jq --argjson r "$row" '. + [$r]' "$file" > "$tmp" 2>/dev/null; then
+      _bypass_audit_preserve_mode "$file" "$tmp"
+      mv "$tmp" "$file" || exit 1
+      # Successful rename: clear the trap so the EXIT path doesn't try
+      # to rm a now-renamed (i.e. nonexistent) tmp path. Also let the
+      # outer caller manage the lock_dir cleanup uniformly.
+      trap - EXIT INT TERM
+      exit 0
+    else
+      rm -f "$tmp"
+      echo "[FAIL] bypass_audit_append: jq failed" >&2
+      trap - EXIT INT TERM
+      exit 1
+    fi
+  ) || rc=1
 
   rmdir "$lock_dir" 2>/dev/null
   return "$rc"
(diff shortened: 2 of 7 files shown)


kraulerson:
## Adversarial review — PR #93

### Verification

- **TDD validated**: checked out PR branch, reverted `scripts/lib/host.sh`, `scripts/lib/bypass-audit.sh`, and `scripts/hooks/bypass-detector.sh` to `main` while keeping new tests. All three test files fail with the expected assertions on pre-fix code (5/5, 5/7, 2/5 fail) and pass cleanly on PR HEAD. TDD is honest.
- **Lints clean**: `lint-counter-antipattern.sh` and `lint-backlog-references.sh` both green on PR branch.
- **Regression suites green**: `test-bypass-audit-lib.sh` (10/10), `test-bypass-audit-integrity.sh` (8/8), `test-bypass-detector.sh` (15/15), `test-init-other-host-attestation.sh` (2/2), `test-check-phase-gate-backstop-attestation.sh` (3/3).
- **`host.sh` fix is correct**: explicit dual-parser branch + named-value stderr + `return 1`. The `then=` variable name is allowed by bash (keyword only in syntactic position) but stylistically iffy; not a blocker.
- **`bypass-detector.sh` fix is correct**: envelope `.session_id` read with empty-string treated as missing; env var falls back; final `unknown` sentinel preserved.

### Findings

#### MAJOR — trap clobbers caller's EXIT handler (`scripts/lib/bypass-audit.sh:109,119,190,200`)

The new `trap '...' EXIT INT TERM` is set at function scope but bash traps are **shell-global**, not function-local. The unconditional `trap - EXIT INT TERM` at lines 119 and 200 then **silently removes any pre-existing trap the caller had set**. Reproducer:

```bash
#!/usr/bin/env bash
trap 'echo CALLER-CLEANUP-RAN' EXIT
source scripts/lib/bypass-audit.sh
bypass_audit_append "$PROJ" "$ROW" >/dev/null
# At exit: CALLER-CLEANUP-RAN never prints; caller's trap was wiped.
```

Confirmed locally: `PRE-CALL TRAP: trap -- 'CALLER_TRAP_FIRED=1; echo CALLER TRAP FIRED' EXIT` / `POST-CALL TRAP:` (empty) — and the caller's trap body never runs at process exit.

No current in-repo caller of `bypass_audit_append`/`bypass_audit_close_pending` sets its own EXIT trap, so this is latent rather than actively breaking behavior today. But this is a **library** function — the next consumer who adds `trap 'cleanup' EXIT` and then calls into bypass-audit will have their cleanup silently destroyed, with no visible failure. That's exactly the silent-bypass anti-pattern this PR was created to eliminate.

Suggested fix: save/restore the caller's trap, or use a subshell for the rename window (`( tmp=$(mktemp ...); trap '...' EXIT INT TERM; ... )`). The subshell approach is the standard idiom and avoids the global-trap footgun entirely. Alternatively, drop the EXIT trap (keep only INT/TERM), since the documented failure modes here are signal-driven; an `EXIT` trap that fires after `return 0` is mostly redundant given the jq-failure branch already `rm -f "$tmp"` explicitly.

#### MINOR — `_bypass_audit_preserve_mode` perpetuates umask-derived modes on fresh files (`scripts/lib/bypass-audit.sh:35-41`, `tests/test-bypass-audit-tmp-hardening.sh:T3`)

`bypass_audit_init` creates the file with `echo "[]" > "$file"`, which inherits the caller's umask (commonly 0022 → 0644). The first `bypass_audit_append` now preserves *that* mode through the rename. So a fresh audit ledger on a default-umask system ends up world-readable (mode 644), and the PR's own T3 test explicitly asserts that this churn-free behavior is correct. Pre-fix, `mv` from `mktemp` downgraded the file to 0600 silently — which was arguably more secure for a governance artifact, even if it ignored operator intent. This trade-off is defensible (operator intent wins), but the audit log carrying user-confirmation rows being world-readable on a default-umask multi-user box is a soft security regression worth a follow-up: have `bypass_audit_init` `chmod 600` after creation, then the preserve-mode helper has a sane baseline to copy from.

#### NIT — `then` as a variable name (`scripts/lib/host.sh:79,87,89,95`)

Bash allows it (keyword only in syntactic position) but several linters flag it. Rename to `then_ts` / `attested_ts` for clarity. Not blocking.

### Scope check

All three assigned findings closed with line-cited fixes; bonus catches (adjacent mktemp scripts; broader CLAUDE_*_ID env-vs-envelope audit) appropriately flagged for follow-up rather than scope-creeped. Comments in `bypass-audit.sh` clearly attribute each hardening to the audit finding.

### Verdict

`minor_concerns` — the trap-clobber is a real defect introduced by this PR, but it is latent (no current caller relies on a pre-existing EXIT trap) and the remaining work is solid. Recommend addressing the trap scope in a follow-up (subshell-wrap or trap save/restore) before this helper accretes more consumers.

kraulerson:
## Verifier follow-up — fix commit 2d5f917

Addresses all three findings from the adversarial review above.

**MAJOR — trap clobbers caller's EXIT handler** (re-labeled major; verifier called it minor on latency grounds but the silent-destruction-of-caller-cleanup is the same class as the silent-bypass bug this PR was created to fix).

Fix: wrapped the `mktemp` → `jq` → `mv` window in a subshell in both `bypass_audit_append` and `bypass_audit_close_pending`. Subshells get their own trap table, so the cleanup trap is contained to the subshell's exit; the caller's `trap 'cleanup' EXIT` now survives untouched. The inner result is propagated via `exit 0/1` and captured as the subshell's rc — no behavioral change to callers of either function.

**MINOR — `bypass_audit_init` perpetuated umask-derived modes on fresh files.**

Fix: `chmod 600 "$file"` immediately after `echo "[]" > "$file"`. Removes the world-readable governance-ledger window on default-umask (0022) boxes and gives `_bypass_audit_preserve_mode` a sane baseline to copy forward on the first append.

**NIT — `then` as a variable name in `scripts/lib/host.sh`.**

Fix: renamed to `then_ts` (4 use sites in `_host_define_other_fallbacks::host_verify_protection`). The `_ts` suffix also makes the unit explicit (epoch seconds).

### TDD evidence

New regression suite `tests/test-bypass-audit-trap-isolation.sh` — 13 assertions across 5 scenarios:
- T1/T2: caller's EXIT trap survives `bypass_audit_append` on both happy + jq-failure paths
- T3/T4: same for `bypass_audit_close_pending`
- T5: `bypass_audit_init` produces mode-600 file even under umask 022

Verified **RED on PR #93 HEAD (df6c208)**: 4 passed, 9 failed (the 9 failures map directly to the MAJOR trap-wipe + MINOR umask findings).
Verified **GREEN on this commit (2d5f917)**: 13 passed, 0 failed.

### Regression suites (all green)
- `test-bypass-audit-lib.sh` 10/10
- `test-bypass-audit-integrity.sh` 8/8
- `test-bypass-audit-tmp-hardening.sh` 7/7
- `test-bypass-audit-schema.sh` 3/3
- `test-bypass-detector.sh` 15/15
- `test-init-other-host-attestation.sh` 2/2
- `test-check-phase-gate-backstop-attestation.sh` 3/3

### Lints (both green)
- `scripts/lint-counter-antipattern.sh` ✓
- `scripts/lint-backlog-references.sh` ✓

### Diff
3 files changed, 306 insertions(+), 34 deletions(-) — most of the bypass-audit churn is comment expansion documenting the subshell-isolation rationale; behavioral churn is small (subshell wrap + one `chmod 600` + variable rename).

PR #96, opened 2026-06-29
feat(ci): two new lints (fix_*() stderr + raw read -rp) + sweep + pre-commit-gate promotion

kraulerson:
## Summary

- New CI lint **fix-functions-stderr**: bans `2>/dev/null` / `2>&-` inside `fix_*()` function bodies (auto-fix functions must surface diagnostics for operators running under `--auto-fix`). PR #92 scrubbed three sites; this lint locks in the precedent.
- New CI lint **raw-read-prompt**: bans bare `read -rp` / `read -p` outside `scripts/lib/helpers.sh`. Bare `read -p` hangs unattended invocations (CI / AI-agent runs) or silently auto-defaults; centralized `prompt_input` / `prompt_yes_no` honor `! -t 0` / `CI=true` / `SOIF_NONINTERACTIVE=true` with safe-default returns.
- Added `prompt_yes_no` to `scripts/lib/helpers.sh` (mirrors the hard-N policy first introduced in `check-phase-gate.sh` per PR #87).
- Promoted both new lints into `scripts/pre-commit-gate.sh` (PreToolUse path + `--terminal-mode` path) alongside the existing two — same `SKIP_LINT=1` escape hatch.
- Swept every in-tree site: 2 fix-functions-stderr sites migrated, 9 raw-read sites migrated to `prompt_yes_no` / `prompt_input`, 15 raw-read sites allowlisted with non-empty reasons (interactive-only wizards: `init.sh`, `intake-wizard.sh`, `check-versions.sh` numeric-choice prompts, `check-phase-gate.sh`'s own wrapper).

## Breaking changes

- **`check-gate.sh` confirmation semantics**: piped `echo y | scripts/check-gate.sh --backfill-host` invocations (no `--yes`) are now **refused on non-TTY stdin** under the hard-N policy. Wrapper scripts that previously confirmed with a piped `echo y` must migrate to passing `--yes` explicitly. `tests/test-check-gate.sh::T2` was updated to lock in the new contract.
- **`intake-wizard.sh` non-TTY refusal** (added in the verifier follow-up commit): the wizard now exits 1 at the interactive mode-selection prompt when stdin is not a TTY unless `SOIF_NONINTERACTIVE=1` is set. Flag-driven passthroughs (`--resume`, `--upgrade-track`, `--upgrade-deployment`, `--to-private-poc`, `--to-sponsored-poc`, `--upgrade-to-production`, `--help`) continue to work in CI.

## Findings closed

(slot 5 — cross-cutting CI lint sweep, no specific finding IDs)

### Pattern 1 — `fix_*()` stderr silencer

| File:line | Fix |
|---|---|
| `scripts/verify-install.sh:778` — `fix_phase_state` | Dropped `2>/dev/null` from `jq` read; malformed `intake-progress.json` now fails loud so the operator can repair it (no half-correct `phase-state.json`). |
| `scripts/verify-install.sh:881` — `fix_intake_suggestions` | Replaced `cp ... 2>/dev/null` with `nullglob`+check so `cp` only fires when source files exist; real `cp` errors surface visibly. |

### Pattern 2 — Raw `read -rp` migrated to `prompt_yes_no` / `prompt_input`

| File:line | Fix |
|---|---|
| `scripts/check-gate.sh:102` | `prompt_yes_no` for `--backfill-host` confirm; CI/non-TTY hard-returns N (prevents auto-Y'ing a manifest mutation). |
| `scripts/process-checklist.sh:446` | `prompt_yes_no` for force-complete confirm (defense-in-depth on top of existing TTY guard). |
| `scripts/process-checklist.sh:1005` | `prompt_yes_no` for reset-process confirm. |
| `scripts/process-checklist.sh:1066` | `prompt_yes_no` for reset-all confirm. |
| `scripts/test-gate.sh:240` | `prompt_yes_no` for `--unrecord-feature` confirm. |
| `scripts/upgrade-project.sh:945` | `prompt_yes_no` for upgrade-plan confirm (preserves "non-interactive mode — proceeding" branch explicitly). |
| `scripts/upgrade-project.sh:1907` | `prompt_yes_no` for auto-install confirm (outer `-t 0` guard already gates this branch). |
| `scripts/verify-install.sh:1333` | `prompt_yes_no` for auto-fix confirm (interactive-only `MODE=interactive` branch). |
| `scripts/check-versions.sh:479` | `prompt_input` for comma-separated selection; CI gets empty default → bounds check below safely skips. |

### Pattern 2 — Raw `read -rp` allowlisted with reason

| File:line | Allowlist reason |
|---|---|
| `scripts/check-phase-gate.sh:50` | Inside its own `prompt_yes_no` wrapper with TTY/CI hard-N guard at lines 40-47; equivalent to `lib/helpers.sh::prompt_yes_no`, retained to avoid a cross-script dependency cycle. |
| `scripts/check-versions.sh:457` | Multi-letter (a/b/c) choice; `prompt_input` / `prompt_yes_no` shape doesn't fit; gated by `-t 0` at line 444. |
| `scripts/intake-wizard.sh:85,88,115,144,146` | Wizard defines its own `prompt_input` / `prompt_choice` / `prompt_with_suggestions` with pause-file semantics; this IS the wizard's centralized prompt helper. |
| `scripts/intake-wizard.sh:1658,1774,1923` | Interactive-only wizard prompts (launch-mode choice, summary confirm, mode selection); wizard is interactive-only by design. |
| `init.sh:396,571,736,781,1955,1977,2042` | Interactive-only init wizard; `NON_INTERACTIVE=true` flag (init.sh:3532) bypasses `collect_inputs_interactive` entirely. The post-1955 attestation prompts also have env-var fast-paths (`REMOTE_URL`, `BRANCH_PROTECTION_ATTESTED`) for non-interactive callers (BL-016). |

## New lint behavior (test results)

**Baseline against `main` BEFORE the sweep:**
```
$ bash scripts/lint-fix-functions-stderr.sh
scripts/verify-install.sh:778: ... fix_phase_state ...
scripts/verify-install.sh:881: ... fix_intake_suggestions ...
2 violation(s) found.

$ bash scripts/lint-raw-read-prompt.sh
... 26 violations across check-gate.sh, check-phase-gate.sh, check-versions.sh,
    intake-wizard.sh (8), process-checklist.sh (3), test-gate.sh, upgrade-project.sh (2),
    verify-install.sh, init.sh (7) ...
26 violation(s) found.
```

**After the sweep (current HEAD):**
```
$ bash scripts/lint-fix-functions-stderr.sh
OK: no fix_*() stderr silencers found.

$ bash scripts/lint-raw-read-prompt.sh
OK: no raw 'read -rp' / 'read -p' calls outside scripts/lib/helpers.sh.

```

## Test plan

**TDD — RED → GREEN:**

`tests/test-lint-fix-functions-stderr.sh` (10 tests):
- RED confirmed: `FATAL: linter not found at .../scripts/lint-fix-functions-stderr.sh` before the linter was written.
- GREEN after lint + sweep: `Results: 10 passed, 0 failed` (covers clean fixture, bare silencer, `2>&-`, non-`fix_` exemption, comment exemption, heredoc-body exemption, allowlist+reason, allowlist no-reason error, MERGE-GATE against `HEAD`, `Reports/` skip).

`tests/test-lint-raw-read-prompt.sh` (10 tests):
- RED confirmed similarly.
- GREEN: `Results: 10 passed, 0 failed` (clean / `read -rp` / `read -p` / helpers.sh exempt / comment / `read -r` no `-p` / allowlist+reason / allowlist no-reason / MERGE-GATE / `Reports/` skip).

`tests/test-pre-commit-gate-lints.sh` extended from 8 → 13 tests:
- T8: PreToolUse blocks fix-functions-stderr violation.
- T9: PreToolUse blocks raw-read-prompt violation.
- T10: `SKIP_LINT=1` bypasses all four lints.
- T11a/T11b: `--terminal-mode` fires both new lints.
- `Results: 13 passed, 0 failed`.

`tests/test-check-gate.sh::T2` — contract update:
- Old T2: piped `echo y` to expect manifest mutation (the bug the migration closes).
- New T2: piped `echo y` WITHOUT `--yes` is correctly refused under non-TTY hard-N policy. `Results: 5/5 passed`.

**Other impacted suites verified passing:**
- `tests/test-lint-counter-antipattern.sh` — 12/12 (regression guard for the existing lint).
- `tests/test-lint-backlog-references.sh` — 15/15.
- `tests/test-process-checklist-classifier.sh` — 12/12.
- `tests/test-process-checklist-auto-advance.sh` — 5/5.
- `tests/test-check-phase-gate.sh` — 8/8.
- `tests/test-check-phase-gate-noninteractive.sh` — 3/3.
- `tests/test-test-gate-counter-sanitizer.sh` — 5/5.
- `tests/test-test-gate-null-handling.sh` — 5/5.
- `tests/test-upgrade-project-atomic.sh` — 4/4.
- `tests/test-intake-wizard-fixes.sh` — 7/7.

## Bonus catches

- Added `prompt_yes_no` to `scripts/lib/helpers.sh` (mirror of the hard-N pattern from `check-phase-gate.sh`) and tightened `prompt_input` to honor the same non-interactive default-return contract — previously `prompt_input` could block forever on a piped stdin under `set -u`.
- Replaced `cp ... 2>/dev/null` in `fix_intake_suggestions` with a proper `nullglob`-based pre-test so the function no longer silently skips a real `cp` failure when source files exist.
- Added a fallback inline `prompt_yes_no` in `scripts/check-gate.sh` for the pre-init migration scenario where `lib/helpers.sh` isn't yet present (with a justified allowlist marker on the fallback's inline `read -rp`).
- `tests/test-check-gate.sh::T2` updated to lock in the new non-TTY hard-N contract — the previous test was inadvertently asserting the bug.

## Worktree note

```
/Users/karl/Documents/Claude Projects/solo-orchestrator/.claude/worktrees/wf_0dedbb55-0f8-5
```

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -69,3 +69,45 @@ jobs:
         env:
           BASE_REF: ${{ github.base_ref || 'main' }}
         run: bash scripts/lint-backlog-references.sh --base "origin/${BASE_REF}" --list || true
+
+  # Cycle-8 wave-3 Slot-5 sibling: ban `2>/dev/null` and `2>&-`
+  # inside any `fix_*()` function body. See
+  # scripts/lint-fix-functions-stderr.sh header for the defect class
+  # (silenced auto-fix stderr makes failures un-actionable for
+  # operators running under --auto-fix). The PR #92 scrub of
+  # fix_framework_clone / fix_framework_manifest / fix_superpowers
+  # established the precedent; this lint is the merge-time backstop.
+  fix-functions-stderr-lint:
+    name: fix-functions-stderr-lint
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Lint fix_*() function stderr silencers
+        run: bash scripts/lint-fix-functions-stderr.sh
+
+      - name: Show PASS/FAIL inventory (on failure or on demand)
+        if: always()
+        run: bash scripts/lint-fix-functions-stderr.sh --list || true
+
+  # Cycle-8 wave-3 Slot-5 sibling: ban raw `read -rp` / `read -p`
+  # outside scripts/lib/helpers.sh — the canonical home for the
+  # prompt_input / prompt_yes_no helpers that handle non-interactive
+  # / CI / no-TTY contexts safely. See
+  # scripts/lint-raw-read-prompt.sh header for the defect class
+  # (bare `read -p` hangs unattended invocations or auto-defaults
+  # without operator confirmation).
+  raw-read-prompt-lint:
+    name: raw-read-prompt-lint
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Lint raw read -rp / read -p sites
+        run: bash scripts/lint-raw-read-prompt.sh
+
+      - name: Show PASS/FAIL inventory (on failure or on demand)
+        if: always()
+        run: bash scripts/lint-raw-read-prompt.sh --list || true
--- a/init.sh
+++ b/init.sh
@@ -393,7 +393,7 @@ collect_project_info() {
     print_warn "For personal projects, Standard track provides external-user readiness without"
     print_warn "enterprise overhead (pen testing, legal review). You can upgrade later."
     local confirm_full
-    read -rp "$(echo -e "  ${BOLD}Continue with Full track? [y/N]${NC}: ")" confirm_full
+    read -rp "$(echo -e "  ${BOLD}Continue with Full track? [y/N]${NC}: ")" confirm_full # lint-raw-read-prompt: allow init.sh interactive-only wizard (NON_INTERACTIVE=true path bypasses collect_inputs_interactive entirely; see init.sh:3532)
     if [[ ! "$confirm_full" =~ ^[Yy] ]]; then
       TRACK=$(prompt_choice "Project track:" "light" "standard" "full")
       TRACK="${TRACK#"${TRACK%%[![:space:]]*}"}"
@@ -568,7 +568,7 @@ collect_project_info() {
   print_info "Directory: $PROJECT_DIR"
   echo ""
 
-  read -rp "$(echo -e "${BOLD}Continue? [Y/n]${NC}: ")" confirm
+  read -rp "$(echo -e "${BOLD}Continue? [Y/n]${NC}: ")" confirm # lint-raw-read-prompt: allow init.sh interactive-only wizard (NON_INTERACTIVE=true path bypasses collect_inputs_interactive entirely; see init.sh:3532)
   if [[ "$confirm" =~ ^[Nn] ]]; then
     echo ""
     local choice
@@ -733,7 +733,7 @@ resolve_and_install_tools() {
   # Confirm — skip prompt when there is nothing to install
   local response="Y"
   if [ "$auto_count" -gt 0 ] || [ "$manual_count" -gt 0 ]; then
-    read -rp "$(echo -e "${YELLOW}▶ ${BOLD}Proceed with this plan? [Y/n]${NC}: ")" response
+    read -rp "$(echo -e "${YELLOW}▶ ${BOLD}Proceed with this plan? [Y/n]${NC}: ")" response # lint-raw-read-prompt: allow init.sh interactive-only tool-install plan; NON_INTERACTIVE path uses AUTO_INSTALL_TOOLS env var rather than this prompt
   fi
   if [[ "$response" =~ ^[Nn] ]]; then
     echo ""
@@ -778,7 +778,7 @@ resolve_and_install_tools() {
         echo ""
         print_info "Default preferences written to: $PROJECT_DIR/.claude/tool-preferences.json"
         print_info "Edit the file, then press Enter to continue."
-        read -rp ""
+        read -rp "" # lint-raw-read-prompt: allow init.sh interactive-only "press Enter to continue" pause after manual tool-preferences edit; non-interactive path skips this whole branch (PROMPT_EDIT only used when -t 0 above)
         # Re-resolve after manual edit
         resolver_output=$("$SCRIPT_DIR/scripts/resolve-tools.sh" \
           --dev-os "$dev_os" \
@@ -1952,7 +1952,7 @@ create_and_protect_remote() {
     if [ -n "${REMOTE_URL:-}" ]; then
       remote_url="$REMOTE_URL"
     else
-      read -rp "Paste the HTTPS clone URL of the remote repo you've created: " remote_url
+      read -rp "Paste the HTTPS clone URL of the remote repo you've created: " remote_url # lint-raw-read-prompt: allow init.sh interactive-only "other host" URL paste; gated by REMOTE_URL env-var fast-path at line 1952 above for non-interactive callers (BL-016)
     fi
     [ -z "$remote_url" ] && { print_fail "Remote URL required for 'other' host"; return 1; }
     git remote add origin "$remote_url"
@@ -1974,7 +1974,7 @@ create_and_protect_remote() {
       attest="yes"
       print_info "Branch protection attested via --branch-protection-attested flag."
     else
-      read -rp "Has branch protection been configured per the above? [type 'yes' to attest]: " attest
+      read -rp "Has branch protection been configured per the above? [type 'yes' to attest]: " attest # lint-raw-read-prompt: allow init.sh interactive-only attestation; gated by BRANCH_PROTECTION_ATTESTED env-var fast-path at line 1973 above for non-interactive callers (BL-016)
     fi
     [ "$attest" != "yes" ] && { print_fail "Attestation required — cannot proceed to Phase 0"; return 1; }
     # Record attestation in process-state.json (BEFORE push — see BL-024 above).
@@ -2039,7 +2039,7 @@ create_and_protect_remote() {
         attest="yes"
         print_info "Branch protection attested via --branch-protection-attested flag."
       else
-        read -rp "Attest that protection will be enforced manually? [type 'yes' to attest]: " attest
+        read -rp "Attest that protection will be enforced manually? [type 'yes' to attest]: " attest # lint-raw-read-prompt: allow init.sh interactive-only attestation (manual-enforcement branch); gated by BRANCH_PROTECTION_ATTESTED env-var fast-path at line 2038 above for non-interactive callers (BL-016)
       fi
       [ "$attest" != "yes" ] && { print_fail "Attestation required — cannot proceed (see $host driver remediation above)"; return 1; }
       # Record attestation with the github_free_tier reason so check-gate.sh
--- a/scripts/check-gate.sh
+++ b/scripts/check-gate.sh
@@ -19,6 +19,22 @@ source "$SCRIPT_DIR/lib/helpers.sh" 2>/dev/null || {
   print_info() { echo "[INFO] $*"; }
   print_warn() { echo "[WARN] $*"; }
   log_line()   { :; }
+  # Wave-3 raw-read sweep: prompt_yes_no fallback honors the same
+  # non-interactive hard-N contract as the helpers.sh version. Reached
+  # only in pre-init migration scenarios where lib/helpers.sh is
+  # absent; behavior must match the canonical helper so the manifest
+  # mutation in --backfill-host below stays consistent.
+  prompt_yes_no() {
+    local message="$1" default_answer="${2:-N}"
+    if [ ! -t 0 ] || [ -n "${CI:-}" ] || [ -n "${SOIF_NONINTERACTIVE:-}" ]; then
+      echo "[WARN] Non-interactive context: skipping prompt (\"$message\") — defaulting to 'N' (caller default '$default_answer' ignored)." >&2
+      return 1
+    fi
+    local reply
+    read -rp "${message}: " reply # lint-raw-read-prompt: allow fallback prompt_yes_no defined inline when lib/helpers.sh is absent (pre-init migration scenario); semantically equivalent to lib/helpers.sh::prompt_yes_no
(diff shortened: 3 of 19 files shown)


kraulerson:
## Adversarial review — PR #96

Two new lints land cleanly (baseline 2 + 26 violations on `main` exactly match what I reproduced; sweep is correct; `tests/test-lint-*.sh` + `tests/test-pre-commit-gate-lints.sh` + `tests/test-check-gate.sh` all green; no unrelated suites regressed vs. `main`). But the PR's own contract — "centralized prompt helpers honor `! -t 0` / `CI` / `SOIF_NONINTERACTIVE`" — has a hole the lint cannot see, plus two smaller concerns.

### 1. **MAJOR — `prompt_install` in `scripts/lib/helpers.sh:292` violates the very contract this PR establishes (and lint cannot catch it because `helpers.sh` is the only exempt file).**

The PR's commit/body both call out: "tightened `prompt_input` to honor the same non-interactive default-return contract — previously `prompt_input` could block forever on a piped stdin under `set -u`." Same file, one helper down (`prompt_install`, line 292) still does:

```bash
read -rp "$(echo -e "  ${BOLD}Install $tool_name now? [Y/n]${NC}: ")" response
if [[ "$response" =~ ^[Nn] ]]; then
  return 1
fi
# falls through to: eval "$install_cmd"
```

No `[ ! -t 0 ]` / `${CI:-}` / `${SOIF_NONINTERACTIVE:-}` guard. On EOF (piped stdin closes), `read` returns nonzero but the script keeps going with `response=""`; `[[ "" =~ ^[Nn] ]]` is false → `eval "$install_cmd"` runs unattended. `prompt_install` is called from 16 sites in `init.sh:117-249` for `sudo apt install -y git`, `sudo apt install -y docker.io && sudo usermod -aG docker $USER`, etc. — i.e. precisely the auto-Y'd side-effectful actions this PR is supposed to prevent. The `scripts/lib/helpers.sh` exemption in `scripts/lint-raw-read-prompt.sh:69` means CI will never flag it. Either (a) move `prompt_install`'s real prompt into `prompt_yes_no` and inherit the hard-N guard, or (b) add the same `! -t 0 / CI / SOIF_NONINTERACTIVE` short-circuit inline. Hiding the hazard inside the one exempt file is the worst outcome for a lint sweep PR.

### 2. **MINOR — `fix_phase_state` "fix" doesn't deliver what its new comment promises (`scripts/verify-install.sh:778-786`).**

The new comment reads: "A malformed `intake-progress.json` should fail loud (the operator then re-runs intake or repairs the file), not silently default to `poc_mode=null` and write a half-correct `phase-state.json`." The actual change:

```bash
pm=$(jq -r '.answers.poc_mode // empty' .claude/intake-progress.json || echo "")
```

The `|| echo ""` still masks jq's exit code; under `set -e`, command-substitution exit status isn't propagated. So when jq fails on malformed JSON, jq's stderr is now visible (good), but `pm` still gets `""`, `poc_value` still defaults to `"null"`, and `phase-state.json` is still written half-correct (exactly what the comment said it would avoid). Either tighten the fix to `|| { print_warn ...; return 1; }`, or rewrite the comment to "surface jq stderr so the operator can diagnose; behavior unchanged."

### 3. **MINOR — `intake-wizard.sh` allowlist reason ("interactive-only by design") is an assertion, not a guard.**

8 sites allowlisted with that justification (`scripts/intake-wizard.sh:85,88,115,144,146,1658,1774,1923`). `grep -nE 'NON_INTERACTIVE|SOIF_NONINTERACTIVE|CI=|! -t 0|\[ -t 0' scripts/intake-wizard.sh` returns nothing. There is no code that refuses to run unattended; the wizard will block-on-stdin / auto-empty-input exactly like the sites that were migrated. If "interactive-only" is the design, add a single `[ ! -t 0 ] && { print_fail "intake-wizard requires a TTY"; exit 1; }` near the top of `main()` so the contract is enforced rather than asserted. Without that, the lint allowlist documents a contract no one is enforcing — same shape as the bug the migration closes for everyone else.

### 4. **NIT — undisclosed caller-facing behavior change.**

`tests/test-check-gate.sh::T2` flipped semantics: `echo y \| check-gate.sh --backfill-host` (no `--yes`) used to write the manifest; now refuses with rc≠0 and no host written. That's the right new behavior, but the PR body doesn't list it as a breaking change for wrapper scripts. A line in the Summary like "Breaking: piped confirmations (`echo y \| ...`) without `--yes` are now refused on non-TTY stdin; migrate wrappers to `--yes`" would save the next caller a debugging session.

### 5. **NIT — regex blind spots in `scripts/lint-raw-read-prompt.sh:74`.**

```
READ_PROMPT_RE='(^|[[:space:]]|;|!|\()read[[:space:]]+-[A-Za-z]*p[A-Za-z]*([[:space:]]|$)'
```

Misses: `read -s -p "x" v` (separate `-s` and `-p` bundles), `read -p"compact" var` (no space between `-p` and value), `read --silent -p ...` (long-form flags). Confirmed by testing the regex on synthetic lines. No in-tree uses of these forms today, but a future drift would slip through. Worth either widening the regex or adding a comment to the lint header noting the supported flag-bundle shapes.

### What I verified

- Baseline `bash scripts/lint-{raw-read-prompt,fix-functions-stderr}.sh` against `origin/main` reproduces 26 + 2 violations (matches body).
- PR HEAD: both lints exit 0, plus `lint-counter-antipattern` + `lint-backlog-references` clean.
- `tests/test-lint-fix-functions-stderr.sh` 10/10, `tests/test-lint-raw-read-prompt.sh` 10/10, `tests/test-pre-commit-gate-lints.sh` 13/13, `tests/test-check-gate.sh` 5/5.
- `tests/test-init-non-interactive.sh` N7 fails on PR (25/26); same failure exists on `origin/main` — pre-existing, not a regression.
- `prompt_install` is called from `init.sh:117,124,126,128,154,157,159,161,176,179,181,183,222,230,245,247,249` — 16 sites including `sudo apt install` calls.

kraulerson:
## Verifier follow-up — fix commit 33e351e5e807bd714f11076c793079ba61171d66

Addresses the 5 findings in the adversarial review above.

### MAJOR — `prompt_install` now honors !TTY / CI / SOIF_NONINTERACTIVE

`scripts/lib/helpers.sh::prompt_install` was the one site the lint cannot catch (`helpers.sh` is exempt by design — it owns the prompt helpers). On EOF / piped stdin, `read -rp` returned nonzero with `response=""`, the `[[ "" =~ ^[Nn] ]]` check evaluated false, and `eval "$install_cmd"` fired UNATTENDED across the 16 `init.sh` call sites (`sudo apt install -y git`, `sudo apt install -y docker.io && sudo usermod -aG docker $USER`, etc.).

Fix: same short-circuit shape as `prompt_yes_no` — emit a `[WARN]` naming the missing tool + install command, return 1 (decline install) without touching `eval`.

New test `tests/test-prompt-install-noninteractive.sh` covers all three trigger contexts:
- **T1** CI=true + piped (non-TTY) stdin
- **T2** stdin closed (EOF, `! -t 0`)
- **T3** SOIF_NONINTERACTIVE=1

All three RED on PR-96 HEAD (current bug fires `eval`, creates marker file) and GREEN after fix. Tests mock `eval` as a function that touches a marker, then assert the marker is absent.

### MINOR — `fix_phase_state` actually fails loud now

`scripts/verify-install.sh:778`. The previous fix kept `|| echo ""` which masks jq's exit code under `set -e` (command-substitution rc isn't propagated). Replaced with an explicit `if ! pm=$(jq ... 2>&1); then ... return 1` that (a) surfaces jq's stderr inside the warning message and (b) refuses to write the half-filled `phase-state.json` — matching what the comment promised. `lint-fix-functions-stderr` still clean.

### MINOR — `intake-wizard.sh` TTY guard

The 8 `read -rp` allowlist entries cited "interactive-only by design" — now actually enforced. Guard placed AFTER all flag-driven passthroughs (`--resume`, `--upgrade-*`, `--to-*-poc`, `--help`) and BEFORE the interactive mode-selection prompt, so scripted CI invocations (`edge-cases-scripts.sh::E19` / `E61`) keep working. `SOIF_NONINTERACTIVE=1` is the harness opt-in.

### NIT — Breaking note added to PR body

`check-gate.sh` no longer accepts `echo y | check-gate.sh --backfill-host` without `--yes`. Documented under a new `## Breaking changes` section in the PR body. The `intake-wizard.sh` non-TTY refusal is documented there too.

### NIT — Lint regex widened

`scripts/lint-raw-read-prompt.sh` `READ_PROMPT_RE` now catches:
- `read -p"compact"` (no space between `-p` and value)
- `read -s -p "..."` / `read -r -p "..."` (separate flag bundles)

Verified via synthetic probe: all 4 problematic shapes match; `while IFS= read -r line` (legitimate, no `-p`) still doesn't. The verifier's `read --silent -p ...` long-form concern is moot — bash's `read` builtin has no long-form options. `tests/test-lint-raw-read-prompt.sh` extended from 10 → 13 tests for the wider shapes.

### Self-verification

```
$ bash scripts/lint-raw-read-prompt.sh && bash scripts/lint-fix-functions-stderr.sh && bash scripts/lint-counter-antipattern.sh && bash scripts/lint-backlog-references.sh
OK: no raw 'read -rp' / 'read -p' calls outside scripts/lib/helpers.sh.
OK: no fix_*() stderr silencers found.

$ bash tests/test-prompt-install-noninteractive.sh        # 3/3 (was 0/3 RED on PR HEAD)
$ bash tests/test-lint-raw-read-prompt.sh                 # 13/13
$ bash tests/test-lint-fix-functions-stderr.sh            # 10/10
$ bash tests/test-check-phase-gate-noninteractive.sh      # 3/3
$ bash tests/test-pre-commit-gate-lints.sh                # 13/13
$ bash tests/test-check-gate.sh                           # 5/5
$ bash tests/test-intake-wizard-fixes.sh                  # 7/7
$ bash tests/test-poc-modes.sh                            # 5/5
```

`edge-cases-scripts.sh::E50` stops with rc=1 — that pre-existing failure exists on PR-96 HEAD and on `origin/main` (same shape as the `test-init-non-interactive.sh::N7` flake the verifier already noted as pre-existing, not a regression).

PR #107, opened 2026-06-29
fix(init): honor NON_INTERACTIVE + AUTO_INSTALL_TOOLS (BL-057)

kraulerson:
## Summary

- **Bug:** `init.sh:736` called `read -rp` unconditionally when the resolved tool plan had any `auto_install` / `manual_install` entries. Under `--non-interactive` (closed stdin + `set -euo pipefail`), `read` returned non-zero and the script terminated silently with `rc=1`.
- **Surfaced by:** Step-5 dogfood validation walker (DOGFOOD-001) — the only bug found across 38 scenarios. Reproducer: `--platform mobile` on a Darwin host without Android Studio (the resolver auto-installs Android Studio at Phase 2).
- **Blast radius:** Today, only `mobile`. Grows with every new `auto_install` entry added to the tool matrix.
- **Fix:** Honors the contract the inline lint-suppression comment already documented — `NON_INTERACTIVE=true` → `response = ${AUTO_INSTALL_TOOLS:-Y}`; the `[Nn]` decline branch now also short-circuits to a "skip + log" path under `NON_INTERACTIVE` instead of dropping into the interactive `prompt_choice` sub-menu (which would EOF-fail under closed stdin).
- **Regression test:** New `tests/test-init-non-interactive-mobile-auto-install.sh` (3 cases — default Y, explicit N, explicit Y). Wired into `tests/full-project-test-suite.sh` as TEST 0c4 per BL-034 invariant.

## Bug source

Step-5 dogfood validation walker (DOGFOOD-001 — see `Reports/2026-06-29-step5-dogfood-validation.md`). The only bug found across 38 scenarios. The inline lint-suppression comment at `init.sh:736` documented the intended bypass but no code in the file actually honored it.

## Repro and fix

**RED on `origin/main`:**

    init.sh --non-interactive --platform mobile --language typescript \
            --track full --deployment personal --gov-mode private_poc \
            --project foo --project-dir <tmp> </dev/null

Result: prints the Tool Installation Plan, then dies silently. No diagnostic. `rc=1`.

**Fix in `init.sh` (`resolve_and_install_tools`):**

```bash
local response="Y"
if [ "$auto_count" -gt 0 ] || [ "$manual_count" -gt 0 ]; then
  if [ "$NON_INTERACTIVE" = true ]; then
    response="${AUTO_INSTALL_TOOLS:-Y}"
  else
    read -rp "..." response   # interactive only; lint-allowlisted
  fi
fi
if [[ "$response" =~ ^[Nn] ]]; then
  if [ "$NON_INTERACTIVE" = true ]; then
    print_info "AUTO_INSTALL_TOOLS=N — skipping tool auto-installation..."
    auto_count=0
    return 0
  fi
  # ... existing interactive prompt_choice sub-menu unchanged ...
fi
```

**GREEN (post-fix, same repro):** `rc=0`; project scaffolded; "Installing tools..." banner fires normally.

**GREEN bypass (`AUTO_INSTALL_TOOLS=N`):** `rc=0`; skip-message logged; install loop NOT emitted.

## Test plan

- [x] **RED reproduces on `origin/main`** — `rc=1`, no diagnostic, after the Tool Installation Plan renders. (Verified by stashing the fix and re-running the dogfood walker invocation.)
- [x] **GREEN on this branch** — `tests/test-init-non-interactive-mobile-auto-install.sh` (3 cases): T1 default → Y → `rc=0`; T2 `AUTO_INSTALL_TOOLS=N` → `rc=0`, no install loop; T3 explicit `AUTO_INSTALL_TOOLS=Y` → round-trips to default.
- [x] **No regression on existing non-interactive scenarios** — `tests/test-init-non-interactive.sh` 28/29 (N7 failure pre-existed on `origin/main`; unrelated).
- [x] **Organizational end-to-end intact** — `tests/test-init-organizational.sh` 2/2.
- [x] **All 5 lints green** — `lint-counter-antipattern`, `lint-backlog-references`, `lint-fix-functions-stderr`, `lint-raw-read-prompt` (allowlist marker still on the interactive branch with refreshed rationale), `lint-fixture-envelopes`.
- [x] **Wired into aggregator** — `tests/full-project-test-suite.sh` TEST 0c4 per BL-034 invariant.

## Worktree note

This PR was developed in an isolated worktree at `.claude/worktrees/agent-ab450bf2dfa0333a4` to avoid coupling to other in-flight work. The BL-057 backlog entry is added in a follow-up amend on this branch once the PR number was known (per `scripts/lint-backlog-references.sh`, a Closed entry must cite a PR# or SHA).

**DO NOT merge.** Per the dogfood-walker spec, this PR is staged for review — the reviewer decides ship vs. revise.

### DIFF
--- a/init.sh
+++ b/init.sh
@@ -730,12 +730,33 @@ resolve_and_install_tools() {
   echo -e "${BOLD}└──────────────────────────────────────────────────────────┘${NC}"
   echo ""
 
-  # Confirm — skip prompt when there is nothing to install
+  # Confirm — skip prompt when there is nothing to install.
+  # BL-057: honor NON_INTERACTIVE + AUTO_INSTALL_TOOLS so the --non-interactive
+  # contract is upheld (closed stdin + `set -euo pipefail` + bare `read -rp`
+  # previously terminated the script silently with rc=1 the moment the resolved
+  # plan included an auto_install or manual_install entry — surfaced as a
+  # Step-5 dogfood DOGFOOD-001 on --platform mobile via Android Studio's
+  # auto_install row).
   local response="Y"
   if [ "$auto_count" -gt 0 ] || [ "$manual_count" -gt 0 ]; then
-    read -rp "$(echo -e "${YELLOW}▶ ${BOLD}Proceed with this plan? [Y/n]${NC}: ")" response # lint-raw-read-prompt: allow init.sh interactive-only tool-install plan; NON_INTERACTIVE path uses AUTO_INSTALL_TOOLS env var rather than this prompt
+    if [ "$NON_INTERACTIVE" = true ]; then
+      response="${AUTO_INSTALL_TOOLS:-Y}"
+    else
+      read -rp "$(echo -e "${YELLOW}▶ ${BOLD}Proceed with this plan? [Y/n]${NC}: ")" response # lint-raw-read-prompt: allow init.sh interactive-only tool-install plan; NON_INTERACTIVE branch above honors AUTO_INSTALL_TOOLS env var
+    fi
   fi
   if [[ "$response" =~ ^[Nn] ]]; then
+    # BL-057: under NON_INTERACTIVE, the user has already expressed an
+    # explicit decline via AUTO_INSTALL_TOOLS=N. Dropping into the
+    # interactive prompt_choice sub-menu (which assumes a TTY) would hang
+    # or surface a confusing EOF diagnostic. Skip the auto-install plan
+    # and let init.sh proceed with whatever is already present — anything
+    # truly required will be re-flagged at the phase-gate check.
+    if [ "$NON_INTERACTIVE" = true ]; then
+      print_info "AUTO_INSTALL_TOOLS=N — skipping tool auto-installation. Missing tools will be re-flagged at the next phase gate."
+      auto_count=0
+      return 0
+    fi
     echo ""
     local config_choice
     config_choice=$(prompt_choice "What would you like to do?" \
--- a/solo-orchestrator-backlog.md
+++ b/solo-orchestrator-backlog.md
@@ -1294,3 +1294,29 @@ PR #87 shipped a minimum-viable WARN that surfaces the risk but does not hard-bl
 **Trigger:** This entry exists as a placeholder in case `wf_c62d9fbe-369` does not close before the next backlog snapshot. If `wf_c62d9fbe-369` ships first, this entry flips to Closed with the PR# citation.
 
 **Related:** `scripts/check-phase-gate.sh:246`; PR #87 (minimum-viable WARN); workflow `wf_c62d9fbe-369` (hard-block in flight); audit code-check-gates-7.
+
+---
+
+## BL-057: init.sh --non-interactive must honor AUTO_INSTALL_TOOLS env var
+
+**Logged:** 2026-06-29
+**Category:** Bug
+**Severity:** High
+**Status:** Closed (2026-06-29, PR #107, commit `a0a4e8d`)
+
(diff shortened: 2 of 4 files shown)


kraulerson:
## Adversarial verification — full-certainty pass

**Verdict:** APPROVE. Test is a real regression guard, not tailored. Fix addresses root cause.

### Method
Ran the new test against (a) the PR branch and (b) the PR branch with **only** `init.sh` reverted to `origin/main` (test file held constant). Then ran two targeted mutations on the PR fix to verify the test catches partial reintroductions.

### Results

| Configuration | T1 | T2 | T3 | Conclusion |
|---|---|---|---|---|
| PR branch (as submitted) | PASS | PASS | PASS | Green baseline confirmed |
| PR branch with `init.sh` reverted to `origin/main` | **FAIL** rc=1 | **FAIL** rc=1 | **FAIL** rc=1 | **Test fails hard on the unfixed code — not tailored** |
| Mutation 1: keep env-aware first branch, remove `[Nn]` decline-branch NON_INTERACTIVE bypass | PASS | **FAIL** (catches exact `prompt_choice: stdin closed (EOF)` diagnostic) | PASS | Negative assertion at `tests/test-init-non-interactive-mobile-auto-install.sh:132-135` catches the partial re-break |
| Mutation 2: remove first env-aware branch only | **FAIL** | **FAIL** | **FAIL** | Original bug surface fully detected |

### Test-integrity findings

1. **RED-on-main case (1)** — the test file's header at `tests/test-init-non-interactive-mobile-auto-install.sh:25-28` documents that the explicit "rc=1, no diagnostic" assertion against `origin/main` was omitted as out of scope for an in-tree fast test. The omission is defensible *and* harmless: I just empirically reproduced the RED-on-main outcome by swapping init.sh in place — all 3 cases fail with rc=1 against the unfixed file, which IS the documented RED behavior. The test catches the bug; it just doesn't pin the SHA being tested against. Nit only.
2. **Tautology check** — no tautologies found. Both mutations I constructed (one removing each half of the fix) are detected by the test. The negative assertions on `"prompt_choice: stdin closed"` and `"What would you like to do?"` (`tests/test-init-non-interactive-mobile-auto-install.sh:132-135`) and on `"[STEP] Installing tools..."` (`tests/test-init-non-interactive-mobile-auto-install.sh:140`) make T2 specifically resistant to a "rc=0 but install loop ran anyway" false-positive.
3. **Positive AND negative assertions** — present in every case. T1: positive=rc=0+plan banner, negative=no skip-message. T2: positive=rc=0+skip-message, negative=no sub-menu, no install-loop banner. T3: positive=rc=0, negative=no skip-message.

### Fix-integrity findings

4. **Root cause addressed** — yes. `init.sh:740-747` replaces the unconditional `read -rp` with an env-aware branch that honors the contract the *pre-existing* `lint-raw-read-prompt: allow` comment already documented (the old comment promised "NON_INTERACTIVE path uses AUTO_INSTALL_TOOLS env var" but no code path actually read the env var — the fix makes that promise true). Decline-branch bypass at `init.sh:748-759` prevents the secondary failure mode where `AUTO_INSTALL_TOOLS=N` would still EOF into `prompt_choice`.
5. **No other unguarded `read -rp` in init.sh** — surveyed all 6 `read -rp` sites; remaining ones (`init.sh:396`, `init.sh:571`, `init.sh:802`, `init.sh:1999`, `init.sh:2022`, `init.sh:2093`) are each gated by env-var fast paths (`REMOTE_URL`, `BRANCH_PROTECTION_ATTESTED`, `PROMPT_EDIT`) or by `NON_INTERACTIVE` checks earlier in the call graph, per their existing allowlist comments.

### Minor observation (nit only — does NOT block)

- **T2 banner-substring brittleness**: `tests/test-init-non-interactive-mobile-auto-install.sh:140` asserts the absence of the literal string `"[STEP] Installing tools..."`. If `print_step` is ever refactored to use a different prefix, this assertion would silently weaken. Consider asserting against the bare `Installing tools...` substring or, better, against an explicit "did not invoke install loop" sentinel. Not load-bearing for this PR — the rc=0 + skip-message positives already cover the contract.

— adversarial verifier (claude-opus-4-7, full-certainty pass)

PR #123, opened 2026-06-30
fix(init): write-perm preflight runs before framework-repo guard (closes BL-041)

kraulerson:
## Summary

- Adds `preflight_target_writable()` helper to `scripts/lib/helpers.sh` — operator-facing write-permission probe that walks to the deepest existing ancestor of the target dir and verifies it is writable.
- Wires the new preflight into `init.sh` to run **BEFORE** the existing `guard_not_in_framework` check, so an operator who points `--project-dir` at an unwritable parent now sees a clear permission error instead of the irrelevant developer-facing framework-repo refusal.
- Restores `tests/edge-cases-pre-init.sh` E8b (was SKIPped since the 2026-06-28 test-integrity audit because no preflight existed and the framework-repo guard masked any real-write probe from inside the framework checkout).
- New regression suite `tests/test-init-write-perm-preflight.sh` pins the layering both ways: preflight wins when target is unwritable, guard wins when preflight passes, neither false-positives outside the framework.
- Aggregator registration: new suite registered in `tests/full-project-test-suite.sh` (TEST 0c5) so a future regression surfaces in the full-suite gate.

## Closes

BL-041 (LB-3 — `init.sh:3494` framework-repo guard layering)

## Test plan

- [x] **RED on pre-fix code.** Ran the new suite against the original ordering: T1 fails with "missing write-permission marker" because `guard_not_in_framework` printed the framework-repo refusal first. T2/T3 pass even pre-fix (T2 doesn't depend on the preflight existing; T3 has no failure path to mask).
- [x] **GREEN after fix.** All 3 tests in `tests/test-init-write-perm-preflight.sh` pass.
- [x] **Mutation experiment.** Manually swapped the two `if !` blocks back to guard-first ordering and re-ran the suite. T1 failed RED again (same shape as the pre-fix run), proving the reorder is load-bearing. Restored the fix and verified T1 GREEN.
- [x] **Failure-path coverage** (both branches exercised).
  - **Success path** = preflight blocks the install (T1 / E8b): rc=1, output contains `Cannot create project directory` / `write permission denied`, output does NOT contain `Refusing to operate inside the Solo Orchestrator framework repo`.
  - **Defense-in-depth path** = preflight passes, guard fires (T2): rc=1, output contains the framework-repo refusal banner.
  - **No-false-positive path** (T3): clean tmpdir outside framework + writable target dir → `--validate-only` exits 0.
- [x] **E8b restored** in `tests/edge-cases-pre-init.sh`. Full-suite run of `tests/edge-cases-pre-init.sh` shows 28 PASS / 3 FAIL / 1 SKIP — E8b is now PASS, and the 3 remaining FAILs (E1×2 + E4) are pre-existing apostrophe/dry-run-name-preservation regressions tracked by BL-040 (sibling slot).
- [x] **Adjacent regressions ruled out.** `tests/test-platform-security-bugs-closer.sh` 7/7 PASS (exercises `guard_not_in_framework` directly + every script that depends on it). `tests/test-init-no-remote-creation.sh` 3/3 PASS (exercises the same init.sh pre-resolution path).
- [x] **All four self-verify lints pass:** `lint-counter-antipattern.sh`, `lint-backlog-references.sh`, `lint-fix-functions-stderr.sh`, `lint-raw-read-prompt.sh`.

### mutation_proofs

| Mutation | Tests catching it | Result |
|---|---|---|
| Swap `preflight_target_writable` and `guard_not_in_framework` so guard fires first again | T1 in `tests/test-init-write-perm-preflight.sh` + E8b in `tests/edge-cases-pre-init.sh` | RED (framework-repo refusal masks the permission error) |

## Coordination note

This PR touches `init.sh` and `tests/edge-cases-pre-init.sh`. **Sibling slot BL-040** ALSO touches both files (different sections: BL-040 edits `dry_run_summary` near init.sh:2781 and E2 in `edge-cases-pre-init.sh`; BL-041 edits the `print_header` block near init.sh:3708 and E8b in `edge-cases-pre-init.sh`). The diffs are in non-overlapping ranges so the merge should be syntactic, but a rebase is likely on whichever lands second.

This PR also touches:
- `scripts/lib/helpers.sh` — adds `preflight_target_writable` between `guard_not_in_framework` and `prompt_choice`. No conflict expected with other slots.
- `tests/full-project-test-suite.sh` — adds TEST 0c5 block after TEST 0c4. No conflict expected.

## Worktree note

Worked in `/Users/karl/Documents/Claude Projects/solo-orchestrator/.claude/worktrees/wf_8bbf1808-203-3` (slot `bl041-framework-repo-guard-layering`).

### DIFF
--- a/init.sh
+++ b/init.sh
@@ -3707,30 +3707,58 @@ HELPEOF
 
   print_header "$VERSION"
 
-  # UAT 2026-04-25 fix (U-N): refuse to scaffold a project inside the
-  # framework repo itself. (--dry-run is allowed for inspection.)
+  # BL-041 (2026-06-30): the write-permission preflight and the
+  # framework-repo guard share a target-dir resolution step. Compute
+  # _early_target once and feed it to BOTH checks, in the order:
+  #   1. preflight_target_writable  — operator-facing
+  #   2. guard_not_in_framework     — developer-facing
   #
-  # security-audits-1 (S3, 2026-04-26 audit sweep): also pass the resolved
-  # --project-dir target so the guard catches a malicious or honest-mistake
-  # caller that supplies --project-dir=$FRAMEWORK_REPO from a benign cwd.
-  # The legacy bl-016 audit row #10 claimed this was already covered; the
-  # cwd-only guard did NOT actually cover it. The target-aware overload
-  # closes the gap.
+  # Rationale: a real operator who points --project-dir at an unwritable
+  # location should get the relevant permission error, not the framework-
+  # repo refusal. The framework-repo refusal still fires (defense-in-
+  # depth) when the preflight passes, and is the actual right answer for
+  # the developer scenario it was designed for.
+  #
+  # The previous ordering also blocked tests/edge-cases-pre-init.sh E8b:
+  # the harness runs init.sh from inside the framework checkout (cwd is
+  # the framework repo), so the cwd check in guard_not_in_framework fired
+  # before any write-permission probe had a chance. Reordering the two
+  # closes that test gap without weakening either guard.
+  #
+  # --dry-run skips BOTH because it never actually writes anything (the
+  # preview is allowed for inspection from any context, including from
+  # inside the framework repo itself).
   if [ "$DRY_RUN" != true ]; then
     # Resolve the effective target dir: explicit --project-dir wins; non-interactive
     # default is $HOME/Code/$ARG_PROJECT; interactive flow resolves later via prompt
     # (in that path the cwd check is the only signal until prompts run, which is
     # the same scope as before).
-    _gnif_target=""
+    _early_target=""
     if [ -n "${ARG_PROJECT_DIR:-}" ]; then
-      _gnif_target="$ARG_PROJECT_DIR"
+      _early_target="$ARG_PROJECT_DIR"
     elif [ "$NON_INTERACTIVE" = true ] && [ -n "${ARG_PROJECT:-}" ]; then
-      _gnif_target="$HOME/Code/$ARG_PROJECT"
+      _early_target="$HOME/Code/$ARG_PROJECT"
     fi
-    if ! guard_not_in_framework "$_gnif_target"; then
+
+    # 1. Operator-facing: write-permission preflight (BL-041 / audit
+    # recommendation C). Must run BEFORE the framework-repo guard so the
+    # permission failure mode is reachable from any cwd. preflight_target_writable
+    # is a no-op when _early_target is empty (interactive flow resolves
+    # target later; the existing downstream existence check at
+    # collect_inputs_non_interactive::project_dir + create_project's
+    # mkdir surface the same failure mode in that path).
+    if ! preflight_target_writable "$_early_target"; then
       exit 1
     fi
-    unset _gnif_target
+
+    # 2. Developer-facing: framework-repo guard. security-audits-1
+    # (S3, 2026-04-26 audit sweep) added the optional target argument so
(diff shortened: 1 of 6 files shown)


kraulerson:
## Adversarial review — APPROVE

Independently verified the BL-041 layering fix. Findings below.

### Failure-path coverage — CONFIRMED
- T1 exercises the **preflight failure path** (chmod 0555 → preflight returns 1 with operator-facing marker; framework-repo refusal banner MUST NOT appear).
- T2 exercises the **guard failure path** when preflight passes (defense-in-depth preserved — framework-repo refusal MUST appear).
- T3 exercises the **no-false-positive path** outside the framework (writable target → `--validate-only` exits 0).
- E8b in `tests/edge-cases-pre-init.sh:506-535` exercises the same matrix from the real edge-cases harness and is now PASS (was SKIP since 2026-06-28). Confirmed full run: 28 PASS / 3 FAIL / 1 SKIP; the 3 FAILs (E1×2, E4) and 1 SKIP (E2) are pre-existing BL-040 territory — unrelated to this PR.

### Independent mutation #1 — preflight bypass
Mutated `init.sh:3750` from `if ! preflight_target_writable "$_early_target"; then` → `if false && ! preflight_target_writable "$_early_target"; then` to force the preflight to never fire.
Result: `tests/test-init-write-perm-preflight.sh` T1 went RED — `missing write-permission marker` and the framework-repo refusal banner appeared in T1's output. T2/T3 stayed GREEN as expected (they don't depend on the preflight existing). Mutation caught.

### Independent mutation #2 — task-rubric-specified reorder
Per task rubric, swapped `guard_not_in_framework` to fire BEFORE `preflight_target_writable` (init.sh:3750–3760). Confirmed T1 went RED with the framework-repo refusal masking the permission error, AND **T3 stayed GREEN** — confirming defense-in-depth boundary: the reorder does NOT affect non-framework invocations, which is the expected behavior. Mutation caught at the layer it should be caught.

### Adjacent suite regression check
- `tests/test-platform-security-bugs-closer.sh` 7/7 PASS (exercises `guard_not_in_framework` directly + T4b which verifies every script that depends on it).
- `tests/test-init-no-remote-creation.sh` 3/3 PASS (exercises init.sh pre-resolution path).

### Minor test-completeness gap (non-blocking)
`scripts/lib/helpers.sh:305` — `[ -n "$target" ] || return 0` (empty-arg → success contract) is not tested. I mutated it to `return 1`: all 3 tests still passed because every test passes an explicit `--project-dir`. In real interactive flow (no `--project-dir`, no `--non-interactive`), `_early_target` stays empty (init.sh:3737–3741), so the mutation would short-circuit init.sh after `print_header` and break all interactive invocations. The documented contract is clear; this is a small contract-coverage gap, not a bug in the shipped code. Not worth blocking on, but worth a test in a future closer.

### Bonus catches (delivered)
- Confirmed: the audit's "recommendation C" preflight did not exist in init.sh before this PR. The fix had to both ADD and ORDER it correctly. The PR body and code comment at `scripts/lib/helpers.sh:276–284` document this honestly.
- Aggregator registration at `tests/full-project-test-suite.sh:235–252` (TEST 0c5) per the test-aggregator wiring invariant pattern (BL-034 closure).

### Verdict
APPROVE. Layering is correct, tests are mutation-proven on the critical path, failure paths are exercised in both directions, defense-in-depth boundary is preserved, no adjacent regressions. The minor test gap above is a nice-to-have, not blocking.

— adversarial verifier (claude-opus-4-7, S3 Wave A / bl041-framework-repo-guard-layering)

PR #124, opened 2026-06-30
fix(init): dry_run_summary echoes project description (closes BL-040)

kraulerson:
## Summary

- `init.sh:dry_run_summary` now prints the operator-supplied `PROJECT_DESCRIPTION` between Name and Platform, so the dry-run preview reflects every input that will be persisted to `PROJECT_INTAKE.md` / manifest.
- Empty descriptions are omitted entirely (no dangling label), and embedded newlines/tabs are collapsed to spaces so a multi-line `--description $'foo\nbar'` value stays on a single summary line.
- `tests/edge-cases-pre-init.sh` E2 drops the SKIP and lands three covering assertions through the non-interactive `--description` flag — the interactive stdin path returns the default for description because `helpers.sh:prompt_input()` short-circuits on a non-TTY stdin, so a stdin-piped malicious value never actually reaches `PROJECT_DESCRIPTION` and the previous assertion would have been vacuous.

## Closes

- BL-040

## Test plan

- [x] TDD red → green for `tests/edge-cases-pre-init.sh` E2 / E2b / E2c.
- [x] Failure-path coverage (BL-066 lesson): T-empty asserts the `Description:` label is omitted when value is empty (no dangling label); T-multiline asserts newline normalization so the column layout doesn't break.
- [x] Mutation experiment: reverting just the `dry_run_summary` diff turns E2 BL-040 + E2c (both assertions) RED while E2b still PASSes (empty case was always correct because there was no Description line at all). Restoring the fix returns all six new assertions to GREEN.
  - RED log: `/private/tmp/claude-501/-Users-karl-Documents-Claude-Projects-solo-orchestrator/7492d236-9edf-4ad1-845b-634f9df45abf/scratchpad/mutation-run.log`
  - GREEN log: `/private/tmp/claude-501/-Users-karl-Documents-Claude-Projects-solo-orchestrator/7492d236-9edf-4ad1-845b-634f9df45abf/scratchpad/green-run-v2.log` (33 PASS / 3 pre-existing FAIL — E1 ×2 and E4 — on origin/main, unrelated to this PR / 1 pre-existing SKIP for E8b waiting on BL-041).
- [x] Belt-and-suspenders: `E2: non-interactive dry-run treats description as literal (no command substitution)` — asserts neither `uid=` nor `root:` leak into the dry-run output even on the non-interactive path.
- [x] All four required lints pass (`scripts/lint-counter-antipattern.sh`, `scripts/lint-backlog-references.sh`, `scripts/lint-fix-functions-stderr.sh`, `scripts/lint-raw-read-prompt.sh`).

## Coordination note

Files touched: `init.sh` (added a 12-line block inside `dry_run_summary` near line 2956) + `tests/edge-cases-pre-init.sh` (E2 / E2b / E2c rewrite).

S3 Wave A Slot 3 (BL-041) ALSO touches `init.sh` but in a different region — the framework-repo guard around line 3494 (vs. `dry_run_summary` near line 2956). Conflicts are extremely unlikely, but a textual rebase may still be needed if the sibling PR introduces nearby line shifts. BL-040 lands first by design (BL-041 verification removes E8b SKIP, which currently sits below the E2 region we modified).

## Worktree note

`pwd`: `/Users/karl/Documents/Claude Projects/solo-orchestrator/.claude/worktrees/wf_8bbf1808-203-2`

### DIFF
--- a/init.sh
+++ b/init.sh
@@ -2956,6 +2956,18 @@ dry_run_summary() {
 
   echo -e "${BOLD}Project:${NC}"
   echo "  Name:      $PROJECT_NAME"
+  # BL-040 (2026-06-30): echo the operator-supplied description so the
+  # dry-run preview reflects every input that will be persisted to
+  # PROJECT_INTAKE.md / manifest. Omit the line entirely when the value
+  # is empty (keeps the summary clean for the "no description provided"
+  # default). Newlines and tabs are collapsed to spaces so a multi-line
+  # value reachable through --description $'foo\nbar' doesn't break the
+  # one-line-per-field column layout below.
+  if [ -n "${PROJECT_DESCRIPTION:-}" ]; then
+    local _summary_desc
+    _summary_desc=$(printf '%s' "$PROJECT_DESCRIPTION" | tr '\n\t' '  ')
+    echo "  Description: $_summary_desc"
+  fi
   echo "  Platform:  $PLATFORM"
   echo "  Track:     $TRACK"
   echo "  Language:  $LANGUAGE"
--- a/tests/edge-cases-pre-init.sh
+++ b/tests/edge-cases-pre-init.sh
@@ -198,32 +198,116 @@ else
   pass "E2: backtick command substitution was NOT executed"
 fi
 
-# Audit finding tests-edge-cases-2: the original block had both branches
-# call pass(), so the assertion added 1 to the count regardless of whether
-# the literal "$(whoami)" appeared in the dry-run output. The audit's
-# recommended option A (require the literal to be echoed back) does not
-# match init.sh's actual dry-run contract — dry_run_summary (init.sh:2781)
-# prints project name, platform, track, language and directory but NOT
-# the description (the description is stored in state/manifest only).
-# So a positive-assertion for $(whoami) in dry-run output would fail not
-# because of an injection regression but because dry-run was never
-# designed to echo description back.
-#
-# Adopt audit recommendation B instead: delete the vacuous literal-text
-# block entirely. Injection-resistance is already covered by the uid=
-# check above (init.sh:154) and the DRY RUN completion check below
-# (init.sh:169). Replace the vacuous PASS with a SKIP that explicitly
-# names why this case isn't asserted, so a future change adding the
-# description to dry-run output surfaces a follow-up to wire a real
-# positive grep here.
-# (Closes audit finding tests-edge-cases-2.)
-skip "E2: dry-run does not echo description back — no surface to assert literal \$(whoami) preservation (see init.sh:2781)"
+# BL-040 (2026-06-30) restored assertion: dry_run_summary now echoes
+# the operator-supplied description, so the literal "$(whoami)" and
+# "`id`" from the malicious input must appear verbatim in the dry-run
+# preview. The original E2 stdin-piped harness CANNOT verify this
+# because scripts/lib/helpers.sh:prompt_input() returns the default
+# (empty string) whenever stdin is not a TTY — the piped malicious
+# description never reaches PROJECT_DESCRIPTION, so a regression that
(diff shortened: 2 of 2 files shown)


kraulerson:
## Adversarial review — PR #124 (BL-040 dry_run_summary description)

**Independent mutation experiment (different from implementer's full-revert)**

Ran two targeted mutations against `init.sh:2966-2970` in the worktree:

1. **Remove `tr '\n\t' '  '` normalization** (keep guard, keep echo):
   - `[FAIL] E2c: multi-line description not collapsed onto a single summary line`
   - `[PASS] E2c: dry-run echoes both halves of multi-line description` (both markers still appear — they just end up on separate lines)
   - Confirms E2c *single-line* assertion is a real guard on the `tr` contract, not tautological.

2. **Remove `if [ -n ... ]` guard** (always echo, empty value):
   - `[FAIL] E2b: empty description should be omitted from summary (found dangling label)`
   - Confirms E2b *omit-when-empty* is a real guard on the operator-friendly branch — the mutation produces the exact 'Description: ' dangling label the test rejects.

Combined with the implementer's full-revert experiment (E2 BL-040 echo goes RED), all three novel assertions have independent, non-vacuous failure-path coverage. The implementer's own honesty about the three belt-and-suspenders assertions (E2 no-substitution, E2b completes, E2c both-halves) being vacuous-on-origin but valid future-regression guards is correct.

**Failure-path coverage (BL-066 lesson): PASS**
- Happy path: description appears verbatim (E2).
- Empty path: label omitted entirely (E2b).
- Multi-line path: newlines collapsed to spaces (E2c).
All three exercise both success AND documented failure modes.

**Scope check (per rubric)** — searched for other summary surfaces:
- `dry_run_summary` (init.sh:2952) — updated (correct target per BL-040 action list).
- Post-init 'Setup Complete' box (init.sh:2778) — shows Project + Location only, no description. BL-040 explicitly scopes to `dry_run_summary`, so this is *not* a scope miss.
- No `verbose_summary` / alternate dry-run summary function exists (grep -n confirms only two `Project:` label sites total).

**Lint check** — all four required lints pass on the current PR state:
- `lint-counter-antipattern.sh`, `lint-backlog-references.sh`, `lint-fix-functions-stderr.sh`, `lint-raw-read-prompt.sh` all OK.

**Bonus catch verification** — the 'stdin-piped harness is structurally vacuous' claim is correct: `scripts/lib/helpers.sh:151` short-circuits `prompt_input` on non-TTY / CI / SOIF_NONINTERACTIVE, returning the default. Piping malicious content over stdin never reaches `PROJECT_DESCRIPTION`. Switching E2 to the `--description` flag is the right fix and gives a real positive assertion.

**Minor nit (not blocking)** — `init.sh:2969` breaks the summary's column layout. All other labels (`Name:`, `Platform:`, `Track:`, `Language:`, `Directory:`) align values at column 14; `Description:` is 12 chars vs the previous max of 10, so its value lands at column 16:
```
  Name:      foo
  Description: hello world      <-- value shifted 2 cols right
  Platform:  web
```
The block comment on line 2963-2965 explicitly cites 'one-line-per-field column layout' as the rationale for `tr` normalization, so the label itself already violating that layout is a small internal contradiction. Could be fixed by widening all label padding (2-space bump), or accepted as-is. Purely cosmetic — no test coverage lost either way.

**Verdict: APPROVE.** Independent mutation experiment confirms the three novel assertions (E2 echo, E2b empty-omit, E2c tr-collapse) each guard a distinct code path; failure-path coverage matches the BL-066 lesson; scope is correctly bounded to `dry_run_summary` per the backlog action list; lints clean; bonus catch (vacuous stdin harness) is a genuine and well-explained find.

— adversarial verifier (claude-opus-4-7, S3 Wave A / bl040-dry-run-summary-description)

PR #125, opened 2026-07-01
perf(lib): split helpers.sh into core + full (BL-046)

kraulerson:
## Summary

- Splits `scripts/lib/helpers.sh` (~360 lines) into `helpers-core.sh` (~316 lines: print_*, prompt_*, log_line, run_with_timeout, guard_not_in_framework) and `helpers-full.sh` (~101 lines: init_log/finalize_log + MCP-detection helpers, transitively loads core).
- `scripts/lib/helpers.sh` remains as a thin backwards-compat shim — every existing caller keeps working unchanged. Twelve short-lived callers migrated to source `helpers-core.sh` directly (check-changelog, check-gate, check-maintenance, check-phase-gate, check-session-state, check-updates, check-versions, pending-approval, process-checklist, resume, test-gate, validate).
- Fast-path dirname via `${BASH_SOURCE[0]%/*}` (no subshell fork) — a naive `$(cd ... && pwd)` shim would have regressed full-caller path by ~4 ms per source on Darwin bash 3.2.
- Idempotent-source sentinels (`_SOIF_HELPERS_*_LOADED`) prevent double-parsing when scripts compose (shim → full → core).
- `init.sh` copies all three files into new projects; `upgrade-project.sh` backfills the two new siblings into pre-BL-046 projects before refreshing helper scripts (otherwise refreshed check-*.sh callers would hit "file not found" when sourcing `helpers-core.sh`).
- `scripts/lint-raw-read-prompt.sh` HELPERS_PATH → HELPERS_PATHS[] so all three helper files stay exempt from the raw-read check.

## Closes

BL-046

## Wall-clock measurement

Method: 500-iteration amortized loop, Darwin bash 3.2, M-series Mac (macOS 25.4). Warmup 3 runs discarded; median of 15 samples.

```
bash -c "for i in $(seq 1 500); do (source scripts/lib/helpers.sh); done"
bash -c "for i in $(seq 1 500); do (source scripts/lib/helpers-core.sh); done"
```

| Target | Total (500x) | Per-source | Delta |
|---|---|---|---|
| helpers.sh (shim -> full -> core) | 550.5 ms median | **1.101 ms** | baseline |
| helpers-core.sh (core only) | 411.4 ms median | **0.823 ms** | **-0.278 ms (-25%)** |

Single-source measurement (60 samples, `bash -c "source <file>"`):

| Target | Median | Min | p90 |
|---|---|---|---|
| bash startup only | 2.18 ms | 1.92 ms | 2.36 ms |
[...]

PR #129, opened 2026-07-01
fix(tests): T5 + T5b in test-bl046-helpers-split now actually test the idempotency guard (closes BL-068)

kraulerson:
## Summary

- PR #125's adversarial verifier found T5 (`:205-231`) and T5b (`:233-252`, executed twice for `helpers-full.sh` and the `helpers.sh` shim) in `tests/test-bl046-helpers-split.sh` were vacuous — mutation experiment (delete each `_SOIF_HELPERS_*_LOADED` guard in turn) left all 8 subtests GREEN. Three idempotency guards had zero regression coverage on `main`.
- **T5** cleared `BOLD=""` and asserted it stayed empty; `bash -c` runs non-TTY so helpers-core.sh's `[ -t 1 ]` ELSE branch reassigns `BOLD=""` on every source. The assertion held for the wrong reason.
- **T5b** compared `first="${sentinel:-}"` to `still="${sentinel:-}"`; the sentinel is assigned to literal `1` unconditionally, so `first==still` regardless of whether the guard fired.
- **Fix**: each rewrite plants an observable marker in a variable each file assigns AFTER its guard. T5 plants `LOG_FILE=/tmp/marker-$$` (wiped to empty by `LOG_FILE=""` at line 57 if the guard doesn't fire). T5b plants `MARKER` in `$dirvar` (`_SOIF_HELPERS_FULL_DIR` / `_SOIF_HELPERS_SHIM_DIR`, both recomputed from `BASH_SOURCE` if the guard doesn't fire).
- Backlog: BL-068 filed AS Closed (this-PR-closes-it — surfaced by PR #125's verifier, remediated here before regression could land). PR# placeholder will be replaced with the real number post-merge (consistent with recent Karl pattern — see git log for BL-053 / BL-046 closer edits).

## Closes

BL-068 (new, filed as part of this PR)

## Test plan

**Red → Green cycle**

Baseline (guards intact, before rewrite): 8 passed / 0 failed.

Rewrite (guards intact): 8 passed / 0 failed.

**Mutation proofs — three independent guard removals, each restored before the next**

1. Removed `helpers-core.sh` guard (lines 18-20):
   ```
   [FAIL] T5 idempotent-source — rc=1; output=second source re-ran LOG_FILE reset (guard failed): LOG_FILE=
   ```
   RED as required. T5b's helpers-full and shim continued to pass, confirming isolation.

2. Restored core guard, removed `helpers-full.sh` guard (lines 21-23):
   ```
   [FAIL] T5b idempotent-source (helpers-full.sh) — rc=1; guard did not fire: _SOIF_HELPERS_FULL_DIR changed from /tmp/bl068-t5b-guard-marker-<pid> to .../scripts/lib
   ```
   RED as required. T5 and T5b-shim continued to pass, confirming isolation.

3. Restored full guard, removed `helpers.sh` shim guard (lines 27-29):
   ```
   [FAIL] T5b idempotent-source (helpers.sh) — rc=1; guard did not fire: _SOIF_HELPERS_SHIM_DIR changed from /tmp/bl068-t5b-guard-marker-<pid> to .../scripts/lib
   ```
   RED as required. T5 and T5b-full continued to pass, confirming isolation.

All three guards restored. Final run: 8 passed / 0 failed. `git diff scripts/lib/` empty at commit time — no source-file changes shipped.

**Failure-path coverage**

Each rewrite plants a marker path via `-$$` (PID) so parallel test runs don't collide, and asserts equality (not just non-empty) against the exact marker. A regression that partially preserves the marker (e.g. keeps the prefix but truncates) still fails RED because the equality check is exact.

**Self-verify lints (all EXIT 0)**

- `bash scripts/lint-counter-antipattern.sh` — OK
- `bash scripts/lint-backlog-references.sh` — OK
- `bash scripts/lint-fix-functions-stderr.sh` — OK
- `bash scripts/lint-raw-read-prompt.sh` — OK
- Skipped: `lint-tests-registered.sh` — BL-067 tracks its runtime issue.

## Coordination note

Files touched:
- `tests/test-bl046-helpers-split.sh` (T5 + T5b rewrites, +57/-21 lines net)
- `solo-orchestrator-backlog.md` (BL-068 entry appended, +25 lines)

No source-file changes. No caller changes. No expected merge conflicts against sibling Wave-C PRs (BL-059 / BL-060 / BL-061 touch different files). If merged after PR #125's follow-up commit that fills in `PR #<pending>` → `PR #125` on BL-046, this PR's BL-068 entry will keep its own `PR #<pending>` placeholder to be filled in post-merge with the real number for THIS PR.

## Worktree note

pwd: `/Users/karl/Documents/Claude Projects/solo-orchestrator/.claude/worktrees/wf_42182049-901-4`

Slot: `bl068-tests-bl046-helpers-split-vacuous`.

### DIFF
--- a/solo-orchestrator-backlog.md
+++ b/solo-orchestrator-backlog.md
@@ -1758,3 +1758,28 @@ time bash scripts/lint-tests-registered.sh
 ```
 
 **Related:** PR #122 / BL-038 (introduced the lint — the invariant is correct, only the implementation shape is slow); `scripts/pre-commit-gate.sh` (the caller that pays the cost per commit); `tests/full-project-test-suite.sh` and the other aggregators the lint walks; BL-034 (test-aggregator wiring invariant — this lint is the enforcement arm of that invariant, so keeping it fast is what keeps the invariant credible).
+
+---
+
+## BL-068: T5 + T5b in `tests/test-bl046-helpers-split.sh` are vacuous — idempotency guards have zero regression coverage
+
+**Logged:** 2026-06-30 (PR #125 verifier finding)
+**Category:** Bug / test integrity
+**Severity:** Medium
+**Status:** Closed (2026-06-30, PR #<pending>, this-PR-closes-it)
+
+**What:** The final two subtests of `tests/test-bl046-helpers-split.sh` — T5 (`:205-231`) and T5b (`:233-252`, executed twice: once for `helpers-full.sh`, once for `helpers.sh`) — are tautological. Both claim to prove the `_SOIF_HELPERS_*_LOADED` idempotency guards in `scripts/lib/helpers-{core,full}.sh` + `scripts/lib/helpers.sh` fire on second source. Neither actually does. Mutation experiment: delete the `if [ -n "${_SOIF_HELPERS_CORE_LOADED:-}" ]; then return 0; fi` guard from `helpers-core.sh` and re-run the suite — **all 8 tests still PASS**. Same result for the guards in `helpers-full.sh` and `helpers.sh`. The three guards have zero regression coverage on main.
+
+**Why they're vacuous:**
+- **T5** clears `BOLD=""` between two sources of `helpers-core.sh`, then asserts `BOLD` is empty after the second source ("proof" the guard short-circuited before the color block). But `bash -c "..."` runs in a non-TTY subshell, so `helpers-core.sh`'s `[ -t 1 ]` check takes the ELSE branch and re-assigns `BOLD=''` unconditionally on every source. The assertion passes for the wrong reason. Removing the guard changes nothing.
+- **T5b** captures `first="${sentinel:-}"` after the first source and `still="${sentinel:-}"` after the second, then asserts `first == still`. Each sentinel is assigned to the literal `1` unconditionally at the top of every helper file, so `first=1, still=1` regardless of whether the guard fired. The assertion holds for any file that assigns `X=1` at all, guard-protected or not.
+
+**Impact:** The guards are the entire perf story of BL-046 (avoiding re-parse of the color block, dirname resolution, and function redefinition on nested composition). Without a regression guard, a future refactor that "cleans up" the `_SOIF_HELPERS_*_LOADED` sentinels lands silently, and every short-lived caller pays double parse cost per source. Given how many `check-*.sh` / `validate.sh` callers source helpers-core.sh, that's a real (if small-per-call) tax that compounds under repeated invocation.
+
+**Fix (this PR):** Rewrite T5 and T5b to plant an OBSERVABLE marker in a variable each file assigns AFTER its guard, then assert the marker survives the second source. Concretely:
+- **T5** plants `LOG_FILE="/tmp/bl068-t5-guard-marker-$$"` between sources. `helpers-core.sh` sets `LOG_FILE=""` at line 57 (after the guard at line 18). Guard fires → return before line 57 → marker preserved. Guard removed → marker wiped to empty. Mutation-verified: removing the core guard flips T5 RED.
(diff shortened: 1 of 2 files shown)


kraulerson:
## Adversarial verification — PR #129 (BL-068)

**Verdict: APPROVE — mutation-proof holds; rewrite is not vacuous.**

### Independent verification performed

**Baseline (guards intact):** 8/8 GREEN. Confirms rewrite doesn't itself regress.

**Independent mutation (different from implementer's — targeted the CONDITION not the guard block).** I mutated `scripts/lib/helpers-full.sh` guard from `if [ -n "${_SOIF_HELPERS_FULL_LOADED:-}" ]; then return 0; fi` → `if false; then return 0; fi` (guard body left present but never fires). Ran the suite:

```
[PASS] T5 idempotent-source: helpers-core.sh guard holds (LOG_FILE marker preserved)
[FAIL] T5b idempotent-source (helpers-full.sh) — rc=1; guard did not fire: _SOIF_HELPERS_FULL_DIR changed from /tmp/bl068-t5b-guard-marker-51790 to .../scripts/lib
[PASS] T5b idempotent-source: helpers.sh guard held (_SOIF_HELPERS_SHIM_DIR marker preserved)
```

Correct: T5b (full) fails RED, T5 and T5b (shim) stay GREEN. Guard restored → 8/8 GREEN. Isolation confirmed under a mutation the implementer did not run.

**Impl-parity mutation (core guard delete, same as impl mutation #1).** Reproduced RED for T5, GREEN for T5b×2. Matches the claim exactly:

```
[FAIL] T5 idempotent-source — rc=1; output=second source re-ran LOG_FILE reset (guard failed): LOG_FILE=
```

### Design audit of the rewrite

- **T5 witness (`LOG_FILE=""` at helpers-core.sh line 57):** unconditional assignment after the guard. Only way marker survives second source is guard short-circuits before line 57. No `[ -t 1 ]` branch dependency, no TTY confound. Sound.
- **T5b witness (`_SOIF_HELPERS_FULL_DIR` / `_SOIF_HELPERS_SHIM_DIR` recomputation from `BASH_SOURCE[0]%/*`):** each file recomputes the dirvar unconditionally after its guard. The planted `MARKER="/tmp/bl068-t5b-guard-marker-$$"` is a path that BASH_SOURCE cannot equal (helpers is not located at that /tmp path), so the equality assertion is a real check, not tautological.
- **Isolation per file:** T5b iterates `helpers-full.sh` and `helpers.sh` separately, each in its own `bash -c` subshell. Second source of shim with intact shim guard returns before re-executing full/core, so shim's dirvar remains preserved even if full's guard were removed — but full's own guard is separately tested by T5b (full). Coverage is per-guard, not accidentally-conflated.
- **PID collision / parallel-run risk:** `$$` inside `bash -c` gives the invoked shell's PID; markers are compared by string, not file existence. No collision surface.
- **Env leak of sentinels into first source:** `_SOIF_HELPERS_*_LOADED` are never `export`-ed, and the aggregator invokes tests via `bash tests/...` (subprocess). Even if a caller set them locally, they wouldn't propagate to the `bash -c` grand-child. Safe.

### Failure-path coverage (BL-066 discipline)

- Success path: guard fires → marker preserved → GREEN. Exercised.
- Failure path: guard removed / guard-condition inverted → witness variable overwritten → RED. Exercised via mutation experiment (three impl mutations + one adversary mutation).

### Backlog hygiene

BL-068 filed AS Closed with `PR #<pending>` placeholder. Matches Karl's recent single-commit-close pattern (BL-053 / BL-046). No lint regression. `lint-tests-registered.sh` skip is properly attributed to BL-067.

### No defects found

Not a defect but worth flagging: T5 is now COUPLED to the specific witness `LOG_FILE=""` on helpers-core.sh line 57. Similarly T5b is coupled to the dirvar recomputation shape. If a future refactor removes those unconditional assignments (e.g., moves LOG_FILE init into `init_log` only), T5/T5b will silently start passing for the wrong reason again. The rewrite comments call this out explicitly (file+line references), so the coupling is visible to future maintainers — mitigation is adequate but not eliminated. Not blocking; would recommend a future consideration to also assert on function-hash / declare-F count post-double-source as a second, structurally-orthogonal witness. Non-blocking.

**Recommendation: MERGE.**

— adversarial verifier (claude-opus-4-7, S3 Wave C / bl068-tests-bl046-helpers-split-vacuous)

PR #207, opened 2026-07-17
fix(gates): close the --no-remote-creation blind spot — pin + push axes (BL-110, BL-116)

kraulerson:
## Findings (the third and fourth corners of the one uncovered flow)

**BL-110:** the `soloFrameworkCommit` birth-stamp lived inside `create_and_protect_remote()` **after** its `--no-remote-creation` early return — every hermetic scaffold (all UAT/CI/agent runs) was born **pin-less**, degrading the Currency System's anchor and `--sync-framework` drift reporting. **BL-116:** the MANDATORY BL-084 push gate was implemented only for `host == "other"` on the premise that first-class hosts are "provably pushed at init" — false for exactly this flow.

## Fix

- **`# BL-110-PIN-UNIVERSAL`**: stamp moved to the universal manifest-seed site in `prepare_initial_state_for_commit` (every path, before the currency stamp; idempotent; remote-path duplicate removed with a pointer comment).
- **`# BL-116-PUSH-GATE-SCOPE`** (excision-safe fence): the first-class exemption is now **earned, not assumed** — the gate is skipped only when `remote_repo_created`+`pushed_initial` are both on record (the on-disk meaning of the premise); `host=other` keeps unconditional gating.

## Proof

- BL-116 RED observed: github/org fixture, no records, no remote → **0 push-gate lines**. Post-fix: FAIL fires; recorded-steps fixture exempt; host=other regression pinned; **fence-excision mutant** restores the old scope and the case regresses (4/4).
- BL-110 RED observed via mutation: HEAD-reverted init scaffolds a real `--no-remote-creation` project with `MUTANT-PIN=ABSENT`; fixed init stamps pin == framework HEAD (`T-scaffold-pin-stamped`, scaffold suite 8/8 with a real init).
- Fixture updates are **honest modeling, not weakening**: bl104/bl102/backstop fixtures now record the init-push steps they always represented — the scope fix correctly began gating "never-pushed" fixtures, which is the fix working.

## Verification

- `test-bl116-push-gate-scope.sh` 4/4 · scaffold suite 8/8 (3 real inits) · bl104 · bl102 · backstop-attestation · bl124 · check-phase-gate · run-lints 11/11

**Stacked on #206** → #205 → #204 → #203. Merge ascending; auto-delete retargets each.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -85,6 +85,7 @@ jobs:
             tests/test-bl118-sast-dom-xss.sh
             tests/test-bl119-stale-editmsg.sh
             tests/test-bl121-cutline-bsd-sed.sh
+            tests/test-bl116-push-gate-scope.sh
             tests/test-bl123-bp-attestation-recovery.sh
             tests/test-bl124-pending-ratchet.sh
             tests/test-bypass-audit-integrity.sh
--- a/Reports/2026-07-13-dogfood-2/REMEDIATION-PROGRESS.md
+++ b/Reports/2026-07-13-dogfood-2/REMEDIATION-PROGRESS.md
@@ -118,4 +118,17 @@
 - **Affected suites (green):** test-check-gate 5/5 · process-checklist ×4 (classifier/auto-advance/subject/reset-phase1) · run-lints 11/11.
 - **Adversarial verify (Opus-tier, laundering-focused brief):** verdict **SHIP conditional on finding A — condition MET pre-push**. A (precondition guard), B (provenance), C (presence-keyed idempotency) all landed in the amended commit, each with its own watched RED. Verifier also independently re-excised the BL-126 fence (excision-safe + fail-closed confirmed) and noted the recorder/steps-writer root-dir coherence is pre-existing (`_require_manifest` anchors cwd).
 
+---
+
+## WP-D2 — BL-110 (Med) pin absent on hermetic path + BL-116 (Med) push gate host-scoped — DONE-PR-open
+
+- **Branch:** `fix/bl110-bl116-noremote-blindspot` (stacked on PR #206). **PR #207**, commit `6e7ecaf`.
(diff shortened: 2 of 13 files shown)


kraulerson:
Adversarial verifier (Opus) verdict: **SHIP — no blockers, no should-fixes.** Highlights: remote-path clobber REFUTED (manifest always exists by the create/protect merge branch, jq-merge preserves the pin; heredoc overwrite unreachable); real hermetic init verified pin==framework HEAD with the manifest delta being exactly +soloFrameworkCommit; BL-116 semantics driven on 6 fixtures incl. legacy shapes (reachable remotes never false-block; truly-legacy projects were already ZDR-blocked, so BL-116 is never their marginal blocker). All suites + 11/11 lints green under the verifier's own runs.

kraulerson:
## CI repair — why this PR went red and what fixed it

**Root cause (this PR):** two unit-lane suites' pass-path fixtures predate the `# BL-116-PUSH-GATE-SCOPE` arm this PR introduces — the gate **correctly** refused them:
- `test-bl073-review-manifest-gate.sh` (4 cases): fixture had no recorded `remote_repo_created`/`pushed_initial`.
- `test-check-phase-gate-poc-block-contract.sh` (T1/T2): same, and its repair had landed one PR too high in the stack (`8f6ea0a`, #208) — a stack-split placement error on my side.

**Fix (`a52506a`, this branch):** both fixtures modernized where the demanding arm lives — bl073 gains the push records (+ `docs/phase-0/` intermediates as forward evidence for #208's BL-114 arm; harmless here, required one merge later); poc-block ported verbatim from `8f6ea0a` (identical content, so later merges stay conflict-free). At this branch: bl073 29/29, poc-block 5/5, bl116 4/4.

**The wider sweep:** the tip PRs were also red — four root causes total, all repaired and propagated down the stack via parent→child merges, so every open PR's merge-ref now carries every fix. Full accounting in the ledger's POST-RUN CI REPAIR entry (rides in #211). **Merge order is unchanged:** #207 → #208 → #209 → #210 → #211.

Owned honestly: none of the five failing suites was in any of my in-run blast-radius batteries, and two failures were lint violations in the run's own new tests that local BSD-grep lint runs never fired on (CI's GNU grep did). Both classes are recorded in the final report's Honest-limits section.

PR #218, opened 2026-07-18
fix(gates): section-bound + literal-tighten the approval placeholder detector (BL-138)

kraulerson:
## What

**Dogfood-3 F-DF3-001:** `validate_approval_fields` refused a **fully-filled** first gate while the walker followed the template's own conventions — the `grep -A 20` window re-anchored on the `| **Gate** |` table row and bled into the template's downstream UAT/Attorney placeholder rows (the same window-bleed class fixed in `_cpg_gate_has_evidence` and `# BL-115-ATTORNEY-ENTRY`; this was the missed third sibling), and the any-bracket predicate flagged `[SIMULATED]` while the bare `YYYY-MM-DD` arm flagged date-format prose.

**Fix:** H2-anchored window that stops at the next `## ` (+20 cap) — table rows can neither anchor nor extend the scan; the shared `$section` also feeds the self-approval check, which now equally reads only its own gate's rows. `# BL-138-APPROVAL-WINDOW` fences the tightened predicate: **template literals only** (`[YYYY-MM-DD]`, `[Name`, `[Attorney`).

## Evidence

- **RED watched:** T1 twin-fixture rc-parity reproduced the walk verbatim (annotated + downstream-placeholder fixture rc=1 with the WARN vs plain twin rc=0); T4 showed prose tripping the bare-date arm.
- **GREEN 5/5** (`tests/test-bl138-approval-window.sh`, BOTH lists): true positives pinned in-section (`[YYYY-MM-DD]` and `[Name…]` in the gate's own cells still flag); false positives pinned out (`[SIMULATED]`, format prose).
- **Mutation with a recorded trap:** the first T5 draft used a placeholder DATE — and the BL-115 date-evidence arm *correctly* rejected it either way, masking the excised mutant. Switched to the `[Name]` reviewer shape, which only the fenced detector rejects; the guardless mutant now positively reaches rc=0. (Oracle-masking-by-sibling-arm, noted in-suite.)
- **Blast radius green:** self-approval + retroactive-approval (both consume the bounded section) · bl104 · poc-block · bl124 · bl102 · zdr · bl073 · date-writeback · trio 17/17.

BL-138 flips to Closed with this PR's merge SHA. (Parallel to #217/BL-137 — independent files, either merge order works.)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01Nez8F7SmVDJ93dRQoeEGvg

PR #220, opened 2026-07-18
fix(phase3): ZAP work dir joins the project tree + VM-mount diagnosis (BL-140)

kraulerson:
## What

**Stacked: #217 → #218 → #219 → #220 (this).**

**Dogfood-3 F-DF3-005:** the ZAP leg's `mktemp` under macOS `$TMPDIR` (`/var/folders/…`) sits outside VM-based Docker runtimes' shared mounts (Colima shares `/Users` only) — the container wrote its report into a bind that never synced back, the driver FAILed a verifiably clean app, and BL-130 correctly refused to attest the FAIL. No path to green on a common runtime.

**Fix:**
- `# BL-140-ZAP-WORKDIR` — the bind host dir is now `$RESULTS_DIR/.zap-work.$$` (the project tree: where the operator works, inside the VM's shared mounts). The original mktemp stays as the excision-fallback, so the fence mutant restores the old behavior exactly.
- `# BL-140-ZAP-MOUNT-HINT` — the no-report FAIL now names the VM-mount gap and the TMPDIR fallback. The FAIL-not-SKIP posture is untouched (an unreadable scan is not a clean scan).

## Evidence
- **RED watched with the root cause on screen:** the new workdir-witness case printed `/var/folders/…/p3-zap-XtH2O4` verbatim.
- `test-bl070-snyk-zap-scanners.sh` **47/47** (three new cases; the mock docker gained an env-gated hostdir witness, inert for existing cases). Dual-fence mutation case covers both fences, positively asserted.
- Driver blast radius green: bl130 4/4 · bl070 license · bl070 threat-model · bl095 9/9.

BL-140 flips to Closed with this PR's merge SHA. A consolidated adversarial verification of the four Dogfood-3 WPs (#217–#220) follows; verdict lands here.

https://claude.ai/code/session_01Nez8F7SmVDJ93dRQoeEGvg


kraulerson:
## Consolidated Dogfood-3 verifier verdict — SHIP-WITH-FIXES, both MUST-FIXes landed (`398dacc`)

**What HELD under attack:** BL-137's `CI=1` spoof is LOW/acceptable (no hook chain reaches check-phase-gate — `CI=1 git commit` is inert end-to-end; strictly dominated by the pre-existing `SOIF_PHASE_GATES=warn` dial; and it self-records). BL-138's loosened predicate is not a false-negative risk (every bracket shape it now misses lives in never-scanned sections; the H2-window-steal is a loud block, not a silent pass). BL-139's explicit-subject paths and `--subject ""` are consistent. BL-140 D2/D3 clean.

**MUST-1 — D1 (High), landed:** `docker -v` rejects a relative host path (rc=125), `RESULTS_DIR` defaults relative, so the **documented bare invocation** broke on every docker runtime — invisible to all 47 fixtures (each passes an absolute `--results-dir`). `# BL-140-ZAP-WORKDIR` now absolutizes against `$PWD`; new `T-zap-workdir-absolute` passes a **relative** `--results-dir` and asserts an absolute `-v` host path (RED watched).

**MUST-2 — D-extra, landed:** the mutation case was a timing Heisenbug (red at normal speed, green under `bash -x`) — same-second per-run timestamp + shared results dir let a no-report sub-run cross-read a prior clean archive. Fixed both: `# BL-140-ARCHIVE-FRESH` `rm -f` before dispatch (product de-flake) + the two mutation sub-runs isolated into their own dirs. Suite **48/48, green 3× consecutive**.

**SHOULD (filed, none blocks the stack):** BL-141 (verify-install/sync must cover the commit-msg hook — the BL-139 backstop is population-conditional for legacy/declined/non-interactive-sync strict-tier projects; Medium), BL-142 (stale `hook-templates.sh` header vs BL-107), BL-143 (self-approval silently skips when the Approver row sits past BL-138's +20 cap; Medium).

Full per-claim disposition in the ledger § DOGFOOD-3 WAVE — CONSOLIDATED ADVERSARIAL VERIFICATION. The whole stack (#217→#220) also carries the `719ddcb` blame-walker CI repair. Merge order: #217 → #218 → #219 → #220.

PR #226, opened 2026-07-19
fix(gates): recover the past-cap Approver row from the walker's own scan (BL-143)

kraulerson:
**Dogfood-3 SHOULD-fix 2 of 3** (wave verifier C3, Medium).

## The defect
`validate_approval_fields`' permissive pre-extraction is `grep -A 20`-capped: an Approver row sitting more than 20 lines past the last gate-name mention yields an empty name, and the `[ -n "$approver_name" ]` guard silently skipped the **entire anti-self-approval control** — zero output — even though the blame walker's own H2-strict scan is uncapped and would locate the row. BL-138's window bounding made the edge reachable; a crafted (or merely verbose) APPROVAL_LOG evades self-approval detection.

## The fix — `# BL-143-PASTCAP-RECOVERY`
The entry's stronger option: when the capped pre-extraction yields no usable name, locate the Approver row with a faithful copy of the walker's own awk contract and take the name from the located line — the full control (per-line blame included) then **runs** instead of warning-and-skipping. Deliberately additive (excision restores the old skip byte-exactly); the truly-absent-row boundary is unchanged and now pinned.

## Discipline
- **Watched-RED:** T1 = a real past-cap self-approval (Approver row 29 lines into its own section, committed by the approver) producing ZERO self-approval output pre-fix.
- **Suite: `tests/test-bl143-pastcap-selfapproval.sh` 5/5, green 3×**, BOTH lists. T2 guards the recovery path against false positives; T3 pins within-cap parity; T4 pins the absent-row status quo; T5 fence-excision mutant with vacuity guards.
- **Blast radius (the BL-138 lesson — every consumer, green, 12 suites):** self-approval · blame-walker · retroactive-approval · bl138 · bl104 · poc-block · gate-integrity trio · check-phase-gate main suite · bl124 · bl102 · date-writeback · backstop-attestation.
- Backlog flips to Closed at merge; consolidated wave verification follows on the open branches.


kraulerson:
Consolidated adversarial verification (Fable-tier, refutation brief across all three wave branches + the full 5-PR `git merge-tree` collision matrix): **verdict SHIP — zero MUST-fixes.** Every demonstrated gap was executed on both trees and proven byte-for-byte pre-existing on main. Two pre-existing SHOULD residuals filed as BL-144/BL-145 (on the #227 branch). Merge note: no required order; tests.yml + full-project-test-suite.sh will show trivial keep-both anchor conflicts between #223/#224/#225/#226 — re-run `lint-tests-registered.sh` after each resolution.

PR #278, opened 2026-07-28
fix(bl112): the SAST receipt must be earned — a >1MB staged blob was silently unscanned

kraulerson:
Closes the **severe** finding from the retroactive review of merged #274, plus its sibling. 11 commits; four adversarial review rounds; five backlog entries filed rather than hidden.

## The severe defect
The emitted SAST arm never passed `--max-target-bytes`, so semgrep's default 1,000,000-byte cap **silently dropped** any larger staged blob — *after* the BL-182 materialization loop, so the partial-coverage guard could not fire. The hook then printed a clean receipt for a file semgrep never opened.

Reproduced through the real emitter, identical content, only padding differing:

```
staged   900,037 bytes -> REFUSED   [BLOCKED] Semgrep detected security issues
staged 1,100,032 bytes -> COMMITTED [OK] semgrep: SAST ran on 1 staged file(s)
```

Semgrep's own stderr said `Targets scanned: 0`. This was the **fifth** instance of this arm's silent-success class (gitlink, `^[0-3]:` stage syntax, PATH_MAX, rename) and the **first producing a positive false attestation** rather than a loud NOTRUN.

## Fixed at two levels, deliberately
- **The instance** — `--max-target-bytes=0` (`# BL-112-MAX-TARGET-BYTES`).
- **The class** — `# BL-112-SCAN-COVERAGE`: the receipt is now an attestation that must be *earned end-to-end*. The arm proves semgrep accepted, parsed and finished its rules over every target it was handed, and **fails closed** to a loud NOTRUN naming the entries it cannot vouch for. This is what stops a sixth trigger, rather than patching the fifth.
- **The sibling** — `# BL-179-TESTARM-FILTER`: `ACMDR` → `ACMDRT`, so a symlink→file refactor can no longer skip the test gate while printing `[OK] BL-125: no source files staged`.

## Review: four rounds, two rejections
`block` (the class was not closed) → `block` (the defect was still live at the tip) → `minor_concerns` → `minor_concerns`. Reviewer-designed mutations were caught in every round, including one the reviewer predicted would survive and did not.

**Five claims were refuted across those rounds, and every one was prose, not code** — four invented universals, then an accurately-quoted upstream release note whose size clause is false on the pinned version. That last one mattered: it sat beside the BL-187 cross-reference, so a maintainer weighing a reinstated cap would have read upstream's authority as saying a cap *cannot* silently drop a staged blob. It can. That is why this PR exists.

## Residue, filed not hidden
**BL-185** nosemgrep is the only unrecorded escape (policy question) · **BL-186** parse-coverage residue · **BL-187** rule-coverage detector is a presence test · **BL-188** · **BL-189** rule-set resolution — a registry `--config` resolving to zero rules is silent and every coverage fact reports complete.

## Verification
`test-bl132-sast-index-scan.sh` **43/43** (was 27) · `test-bl125-commit-test-exec.sh` **18/18** (was 16, and its first-ever type-change coverage) · `test-bl131` 17/17 · `test-bl118` 6/6 · lints green. Watched-RED on every new assertion; mutation proofs both directions on the flag, the class guard, the fail-closed parse, and the filter.

**TL;DR (plain English):** The security scanner that checks your code before saving was silently ignoring any file over 1 MB — and then reporting "all clear" for a file it never looked at. So a large file containing a real vulnerability sailed through with a clean bill of health. This fixes that, and more importantly makes the "all clear" message something the tool has to *earn*: it now proves it actually examined everything it was given, and says so loudly when it can't. Four rounds of hostile review rejected two earlier attempts before this one held.


kraulerson:
**Disposition decided 2026-07-29: this PR stays OPEN, deliberately, and must NOT be merged.**

#280 (merged, `8f36382`) was the portable half carved out of this stack. What remains here splits
cleanly in two.

**Do not merge — one part is known-unsound.** `# BL-186-PARSE-COVERAGE` (5 sites) is built on
semgrep's `Parsed lines` number, which **BL-192 proved reports `~100.0%` for a file the parser never
decoded**. Merging it would ship a positive false attestation — an `[OK] semgrep: SAST ran …` receipt
printed over an unscanned sink. That clause is abandoned on the record, not pending.

**Do not close yet — two parts are still owed.** Measured against `origin/main` today:

- **Four backlog entries exist only here:** BL-185, BL-186, BL-188, BL-189.
- **Five test cases exist only here** (`tests/test-bl132-sast-index-scan.sh`):
  `T-utf16-parse-drop-no-receipt`, `T-mutation-parse-coverage`, `T-parse-coverage-fails-closed`,
  `T-parse-threshold-exact`, `T-mutation-parse-threshold`.

Those five are **correct code against a broken instrument** — the logic is sound; only the thing they
interrogate is untrustworthy. The last two matter disproportionately: they exist because a `-ge 100`
threshold was left unpinned once and a reviewer's `-ge 100` → `-ge 99` mutant survived the entire
PR-blocking set.

**The condition for closing this PR:** BL-198 WP4 lands — i.e. those five cases are restored in shape
with the predicate re-aimed off `Parsed lines` onto a byte-level decode check that depends on nothing
semgrep reports. Until then this PR is the only copy, and closing it first is how they get forgotten.

**If anyone does rebase this:** `## BL-187:` must be **DELETED**, not merged. #280 already landed its
own copy; a naive merge yields two divergent entries under one heading, and no lint catches that.

See BL-192 (diagnosis + the no-pin decision) and BL-198 (the implementation plan, which names these
five cases as WP4).

kraulerson:
**Converted to draft 2026-07-29** — still open, still must not be merged, now also not presenting as
a merge candidate. Unchanged: the close condition is BL-198 WP4 (see the previous comment).

**Its red `unit-shard (sast)` is expected and needs no action.** This branch predates today's fixes,
so it is reproducing the exact defects that have since been diagnosed and merged:

| failing case | cause | fixed on `main` by |
|---|---|---|
| `T-parse-coverage-no-cry-wolf`, `T-empty-target-receipt`, `T-mutation-typechange-filter` | **BL-193** — semgrep writes its scan-status banner to **stdout** on the GitHub runner and **stderr** on macOS; the coverage guard hard-coded stderr, so every clean commit forfeited its receipt | #280 (`# BL-193-STATUS-STREAM`) |
| `T-hook-references-ruleset`, `T-hook-carries-domxss-config` | **BL-183** — `grep -v … \| grep -q` under `pipefail` inverts: the consumer exits on match, the producer takes EPIPE, and a config that IS present reads as ABSENT | #280 (`# BL-183-NO-SIGPIPE`) |

`grep -c 'BL-193-STATUS-STREAM'` → **0** here, **1** on `main`. (`T-parse-coverage-no-cry-wolf` is the
pre-rename name of `T-coverage-no-cry-wolf`, which survived on `main`.)

**Deliberately not rebased.** Rebasing would pull those fixes in and destroy the pre-fix snapshot this
branch is being preserved for — the five withdrawn cases are *correct code against a broken
instrument*, and WP4 needs to see them as they were written. The branch is also `DIRTY` against `main`
after four merges; that is expected and irrelevant to a PR that will never merge.

Nothing here changes the inventory: four backlog entries (BL-185/186/188/189) and five test cases
still exist only on this branch.

kraulerson:
**Close condition satisfied — closing (not merging), branch retained.**

The condition recorded here (BL-198's restore of the five withdrawn cases) landed on `main` in PR #287 (`365df73`): `T-utf16-parse-drop-no-receipt`, `T-mutation-parse-coverage`, `T-parse-coverage-fails-closed`, `T-parse-threshold-exact`, `T-mutation-parse-threshold` are restored **re-aimed** per the BL-198 v2 plan — predicates read the transcode (bytes the hook holds), not semgrep's self-report, and the UTF-16 sink rows flip from 'receipt forfeited' to the stronger 'sink CAUGHT, commit BLOCKED'. The withdrawn originals remain visible at `e87dbd3` on this branch.

The branch is deliberately **not deleted**: it stays the only copy of backlog entries BL-185/186/188/189 and the pre-fix red-shard snapshot. Filing those four entries on `main` (or superseding them) is the recorded residual before this branch can ever be pruned.

PR #293, opened 2026-07-31
fix(bl200): the token-stream-break detector — forfeit-only, canary-pinned; the receipt names its scanner

kraulerson:
## What

BL-200 — the token-stream-break blind spot: an ordinary ASCII source file with a syntax error beside a sink passes every byte-level check the hook has (BL-198 included) and semgrep misses the sink **deterministically** (reproduced first: the same staged `p.innerHTML = window.name` is rc=1/1 finding alone, rc=0/0 findings with `function ((( broken $$$` beside it).

The emitted hook's semgrep invocation gains `--verbose` for one line of output — `[WARN] Syntax error at line <target>:N` — and a third coverage conjunct (`# BL-200-SYNTAX-BREAK`) joins selection and rule-timeout: three facts for three pipeline stages. Seeing the warning **forfeits the `[OK]` receipt** (warn, never block — the BL-192 decision blocks keep report-dependent instruments out of the blocking path); a **respelled warning degrades to today's receipt**, never to a false one. BL-201's deferred receipt line ships here additively: every `[OK]`/`[BLOCKED]`/tool-failed report now prints `scanner: semgrep <version>` (`# BL-200-SCANNER-VERSION`).

## The admissibility terms, met

- **Forfeit-only, proven**: the conjunct is a pure AND-term; every detector failure mode sanitizes to "no break seen" (today's behavior); the verdict arm is echo-only. The reviewer attacked this property directly and found no constructible grant, block, or receipt-on-detector-failure.
- **Canary-pinned**: `tests/test-bl200-syntax-canary.sh`, **network-free** via a measured prefilter fact (semgrep only parses files some rule's literal admits — the probe rule shares the fixture's `innerHTML` literal, which is also why the detector covers exactly the threat model: a file hiding a sink textually contains it). Pins the exact spelling (C1), that the blind spot is still rc=0 (C2, with a good-news failure message), a same-prefilter valid control (C3, non-vacuity), the exactly-once header under `--verbose` (C4), and the hook-grep/canary lockstep verbatim (C5). **Verified 4/4-green against semgrep 1.157.0 (host) AND 1.172.0 (the CI shard's version, scratch venv)** — the first-contact risk is retired.
- **Costs weighed, measured**: `--verbose` is free (2.14s vs 2.15s scan wall); the version capture costs ~0.7s per semgrep-path commit; the rc≥2 stderr dump is kept WHOLE under verbose (truncation is BL-197's class).
- **`Partially analyzed…`** confirmed printing on every verbose scan — the entry's do-not-anchor trap, held.

## Proofs

- Detector suite `tests/test-bl200-syntax-detector.sh`: **RED 0/7 pre-fix → GREEN 9/9 final**, with in-suite mutation cases proving `--verbose` and the syntax conjunct each carry their own weight, stub cases with their own controls (respell degrades / exact forfeits; version-capture failure prints `(version unknown)` and never costs a receipt), the anchor's **width** pinned (T-stub-anchor-indent — the reviewer's round-1 `^`-drop mutant now fails exactly that case, doubly with canary C5), and the **both-streams read** pinned (T-stub-warn-on-stdout — the confirm round's mutant C now fails exactly that case; BL-193 measured semgrep's stream routing as frontend-dependent).
- Two bl132 boundary cases re-pinned as **measured disjunctions** (`7b4b4a6`): the detector *narrowed* two documented boundaries in the safe direction — the zero-ASCII UTF-16 residue and the transcode-fence excision both now forfeit the receipt on a warning-emitting semgrep instead of buying a false `[OK]`. Hard pins (no-transcode/no-block; COMMITTED-vs-REFUSED) retained; a third, unnamed state still fails.
- Suites: bl200-detector 8/8, canary 5/5 (×2 semgrep versions), bl147 63/0 (policy derivation + CI flag parity untouched), bl112 13/0, bl118 7/0, bl125 22/0, bl131 18/0, bl132 56/0; lints 11/11 (attested post-commit). Emitted hook render-diffed against main: executable-line changes are exactly the claimed set; `bash -n` clean.
- Both new suites registered in the aggregator + the tests.yml unit list, **pinned to the sast shard**; no second `tests=(` token introduced (the `_build_unit_list_set` trap respected); `lint-tests-registered` green.

## Docs

`docs/platform-modules/web.md`'s SAST narrative brought to the five-precondition truth — which also repaired **pre-existing BL-198 drift** (PR #287 never updated its decode-gap paragraph; it still described the transcode fix as unbuilt). Backlog: **BL-200 Closed** (`a969686`, re-pins `7b4b4a6`), **BL-201** gains the landed-deferral UPDATE.

## Review (standing gate: adversarial pr-review BEFORE PR creation)

Round 1 verdict **minor_concerns** (non-blocking): every load-bearing claim independently reproduced; one genuine refutation (RF-1 — the anchor comment overclaimed; echoed continuation lines can inflate an already-forfeiting count, while a CLEAN file's content provably never reaches the stream) plus the surviving `^`-drop mutant (R-BL200-1). Both fixed in `aaf91c3` and re-proved. Confirm round on the delta: **minor_concerns, "clear to open the PR"** — both kills verified doubly by the reviewer re-running its own sabotage; its one new note-level survivor (mutant C, the stream-source swap — bounded to under-detection, the accepted degrade class) was killed same-day by its own prescribed pin (`T-stub-warn-on-stdout`, kill verified: the mutant fails exactly that case). R-BL200-3 (echo can mis-label a forfeit's diagnosis — still forfeit-only in every path) and the measured cost ledger (0.7s version capture per semgrep-path commit; `--verbose` free) recorded on the entry.

Named residues (accepted by the entry's own terms): the detector's report-dependence (respell ⇒ under-detect ⇒ pre-BL-200 receipt; canary-guarded in this repo's lane) and the prefilter bound (a sinkless broken file may pass unwarned — it has nothing to hide).


kraulerson:
CI round: the sast shard's single failure was a portability bug in the test helper, not the product — `_mut_n` passed the `--verbose \` literal via `awk -v`, which processes escape sequences (BSD awk keeps the lone trailing backslash, mawk/gawk mangle it, so the exactly-once probe misfired as MIS-TARGETED on a correct hook). Fixed by passing literals via `ENVIRON` (byte-literal on every awk). Note the shard's other first-contact result: the BL-200 canary passed 5/5 on CI's semgrep 1.172.0, and bl132/bl131/bl118 were green with only the two known environment-conditional skips.

PR #386, opened 2026-09-13
fix(tests): stop four suites inheriting the caller's stdin

Asborien:
`tests/full-project-test-suite.sh` — which `CONTRIBUTING.md` tells every contributor to run to validate a checkout — **never terminates** when stdin is an open pipe or socket rather than a terminal or `/dev/null`.

## Measured on `main`

| stdin | result |
|---|---|
| `/dev/null` | rc=0 in **1 second**, 26 cases |
| open pipe (`< <(sleep 300)`) | **still hung at a 60s timeout**, 15 cases emitted |

It blocks at case 16, `T-bl203-session-check-null-safe`.

## Cause

`tests/test-intake-wizard-fixes.sh:618` runs `OUT=$( cd "$D" && bash "$SESSCHECK" 2>&1 )` with **no stdin redirect**, so the child inherits the caller's. `$SESSCHECK` is `scripts/session-test-gate-check.sh`, a SessionStart hook whose envelope read at line 27 is `ENVELOPE=$(cat 2>/dev/null || echo "")`. A bare `cat` never reaches EOF on a pipe that stays open.

Process sample confirms the stack: `command_substitute → read_comsub → zread → read()`.

It presumably passes in GitHub Actions only because the runner supplies `/dev/null`.

## The fix, and why here rather than in the hook

Redirects stdin at **twelve hook invocations across four suites**, rather than hardening `session-test-gate-check.sh`. The hook's read is correct for its real caller — Claude Code passes a SessionStart envelope on stdin — so teaching it to tolerate an open stdin would break the thing that actually uses it.

## Evidence

- `tests/test-bl276-stdin-hang.sh` — **10 passed / 0 failed**, timeout-bounded so the regression suite itself can never hang
- the original hang case now completes in **2 seconds with all 26 cases**, rc=0
- registered in `tests/full-project-test-suite.sh` and the canonical `tests=( )` array

## Why this one first

This is the defect that blocks the evidence `CONTRIBUTING.md` asks for. I have **not** been able to complete a clean baseline run on pristine `main` — two attempts hung, at 26 and 74 minutes. That is this defect obstructing its own diagnosis, and it is disclosed rather than worked around.


kraulerson:

PR #387, opened 2026-09-13
fix(verify-install): two auto-fixers that could never run

Asborien:
Two `verify-install.sh` auto-fixers that could never run. Both are registered with `register_fixable`, presented to the operator as auto-fixable, and dispatched by `run_remediation` to a function that cannot succeed.

## Arm 2 — a CLI verb that does not exist

The shipped command was `claude plugins add superpowers`. Measured on Claude Code 2.1.269:

```
$ claude plugin add superpowers
error: unknown command 'add'          rc=1
$ claude plugin --help | grep install
  install|i [options] <plugin>
```

**`add` has never been a subcommand.** This fixer could not work on any host, scaffolded or adopted. The fix also marketplace-qualifies the name, because the detector immediately above keys on `.enabledPlugins["superpowers@claude-plugins-official"]`, and `docs/cli-setup-addendum.md` documents a second marketplace for this plugin — so a bare name can install under a key the detector does not read.

`--yes` is deliberately not passed: a fixer must not silently execute a marketplace-declared command on the operator's behalf.

## Arm 1 — an unsatisfiable precondition

`has_context()` is unsatisfiable on an **adopted** project. Its only prior sources were `.claude/tool-preferences.json` — the file `fix_tool_prefs` exists to CREATE, so it cannot be its own precondition — and `grep 'Platform:'`-style anchors emitted by a scaffolded `CLAUDE.md` and absent from a project that kept its own.

Measured on an adopted fixture: three `--auto-fix` passes, `Could not fix: tool-preferences.json missing` each time, file still absent. All five `has_context()` callers took the wrong branch.

The fix reads platform/language/track from the state files already opened by that function, mirroring the resolution chain `scripts/lib/plan-staging.sh` already implements. Empty values must not win, because adoption records them as `""`.

## Evidence

| | |
|---|---|
| GREEN | 16/0 on macOS bash 5.3 and 3.2.57; 15/0/1 on ubuntu:24.04 where the live-CLI case SKIPS loudly |
| RED against `ceb450e1` | 3/13 |
| mutants | six, each with a named killer |

Test-first: the suite commit precedes both fix commits. Registration verified by negative control — removing the `tests.yml` line flips the lint to `not-in-unit-lane`.

## Disclosed

Two comment claims are wrong and were carried verbatim rather than diverging from the adopting project that already runs this code: `--scope user` is described as "git's default" when `--scope` is a `claude plugin install` flag, and a load order is said to mirror "DEPLOYMENT above" when DEPLOYMENT is below in file order. Happy to correct in place if preferred.

One entry covers both arms on the grounds that they are the same defect class, with `## BL-256:` as precedent. They are separate commits, so arm 2 can land alone.


kraulerson:

One change on the way in: **BL-260 → BL-284** (the number was taken on `main` before this landed). Markers, the test file and both registrations were renamed to match; your prose is unchanged. The `claude plugin install` correction was reviewed as a supply-chain surface and kept — it targets exactly the id the repo already documents.

PR #388, opened 2026-09-13
fix(tdd-gate): resolve the branch base from the project's own trunk

Asborien:
The `commit-msg` TDD ordering gate — which `verify-install.sh` calls *"post-BL-139 the **ONLY** terminal-path feat/Build-Loop gate"* — is installed and **never fires** on any repository whose trunk is not named `main`.

## The defect

`_tdd_triggers` declines when a test rode earlier on the branch, and resolves the base as the literal `main`:

```sh
if git rev-parse --verify --quiet main       ; then base="main"
elif git rev-parse --verify --quiet origin/main ; then base="origin/main"
fi
```

On a project whose integration branch is not `main`, the range stops meaning "this branch" and becomes the whole divergence between `main` and the real trunk. The earlier-test count is then almost always non-zero and the classifier returns EXEMPT on **every** commit.

**It fails OPEN**, which is why it matters more than a wrong number. An *unresolvable* base already skips the axis and falls through to fire — correct, and unchanged. A resolvable-but-WRONG base is the dangerous case, and it is the common one, because `main` usually exists even where it is not the trunk.

Measured on a real repo with trunk `preview`: 152 test-shaped files in `origin/main...origin/preview`, so the gate returned 1 before it ever looked at the staged set.

## The fix

Read an explicit `integration_branch` key from `.claude/manifest.json`; an absent key resolves to today's literal.

`origin/HEAD` was considered and rejected: it is local config, so a governance gate keyed on it is configurable by the thing it governs.

## Evidence

| | |
|---|---|
| GREEN | 15/0 |
| RED against the unfixed gate | 4/11 |
| mutants | 3, each with a named killer |

**The compatibility guarantee is proven, not asserted.** Cases `b1`–`b5` assert **byte-identical output AND exit code, with and without the fix**, across: no manifest at all, a manifest with no key, `integration_branch: null`, empty string, and an unresolvable base. `MP2` is the mutant that would change a keyless project's outcome, and `b2` kills it.

## Stated limit — please read before merging

**This ships a READER with no WRITER.** Nothing in the framework writes `integration_branch`, so on every project that exists today the absent key still resolves to `main` and the gate stays inert. This does not close the keyless case; it makes closing it possible.

Always-firing on an absent key is not available either — it would false-block every keyless main-trunk project. What this guarantees is byte-identity for existing projects and no NEW permissive resolution.


kraulerson:

Two changes on the way in: **BL-261 → BL-286**, and — from the adversarial review — the base is now resolved through `refs/heads/`/`refs/remotes/origin/` only (a tag or revspec in the key no longer resolves), and when the key is what exempts a commit the gate now says so and names the base. Your own comment anticipated the read-side limit; this makes it loud rather than silent. A3/A4 pin both.

PR #389, opened 2026-09-13
fix(intake): rename a jq parameter that shadows a reserved keyword

Asborien:
## This is YOUR bug, not a discovery

`## BUG-010:` defect (3), logged 2026-09-01, already names this: `def row(label; val)`, `label` as a reserved keyword, and the prescribed fix. This PR implements it and claims no credit for finding it.

## The defect

```
$ jq -n 'def row(label; val): "|";'
jq: error: syntax error, unexpected label, expecting IDENT or BINDING     rc=3
```

`label` is a jq language construct (`label $out | ... | break $out`). The program does not compile, so the appendix writer emits nothing and the intake context table renders empty while the rest of the document looks correct. jq 1.8.2; corroborated against the jq language reference.

One-word change: `label` → `lbl`.

## Evidence

| | |
|---|---|
| GREEN | 9/0 |
| RED against `ceb450e1` | 2/7 |

Registered in the **PR-blocking `unit-shard` lane**, so a future breakage of this jq program is caught by test rather than at the call site.

## Disclosed

**Only the first of the two halves you prescribed is implemented.** BUG-010 also says *"check jq's exit status rather than reading its stdout"*. That is not done here, and is recorded as an open residual rather than dropped.

The residual currently describes the wrong layer — it talks about `render_intake_file || true` at the call site, whereas your prescription is about the jq invocation *inside* it. Different repairs. Happy to reword.

Also: `## BUG-010:` has no body in the repository, so its prescription is quoted from the summary line at `solo-orchestrator-backlog.md:14258` and could not be verified in full. And no lint can see a dangling `BUG-` anchor — both backlog lints key on `BL-[0-9]+`, so the namespace is outside their reach.


kraulerson:

PR #390, opened 2026-09-13
fix(intake): keep a paused section out of completed_sections

Asborien:
Every `run_section_N` ends in an unconditional `save_section`. Once the pause sentinel is set the prompt helpers return empty and the collection loops break — but that call **still fires**, filing the section under `completed_sections` with nothing in it. `is_section_complete` then makes `--resume` skip it for good.

So a paused section is recorded as finished, and the operator never sees it again.

## The fix

A guard in `save_section`: if the pause sentinel exists, render what was answered, say so, and return without filing the section complete.

## Evidence

| | |
|---|---|
| GREEN | 10/0 |
| RED against `ceb450e1` | 2/8 |
| mutants | MP1 and MP2, both killed by R1 |

`E1` drives the **real `--resume`** end to end: a section paused on one run is asked again on the next.

## Disclosed

**The commit message's third justification misdescribes the code it deletes.** It says "the two sections it named as exceptions carry no prompts at all". The deleted block special-cased exactly **one** section, 11.5, and that section has **nine** prompts. The sentence is wrong and it is not needed — the behavioural argument beside it carries the change alone: with the write gone, `last_section` is 11, `start_section` is 12, and 115, 12 and 13 all survive the `-lt` test.

**A residual, open and out of scope**, documented in the entry: the same arithmetic strands the wizard with no pause at all. Completing 11.5 *cleanly* and being interrupted by anything — Ctrl-C, a closed terminal, a crash — resumes at 116 and silently skips sections 12 and 13 while printing "Intake Complete!". The closing banner then tells the operator to paste a Section 13 prompt that never ran. That is pre-existing, untouched by this fix, and wants its own entry; the durable repair is to skip on POSITION in the ordered list rather than on integer value.


Asborien:
Force-pushed once to re-author the tip commit. A test lab had left a placeholder identity in my clone's config, so the last commit carried a wrong author and Signed-off-by. The tree is byte-identical; only the author and sign-off lines changed.

kraulerson:

PR #391, opened 2026-09-13
fix(intake): re-ask on a bare '?' instead of saving it as the answer

Asborien:
`prompt_input` — the wizard's central prompt helper, behind **81 prompts** — had no handling for `?`, the wizard's own documented help key. The banner says *"Type '?' at prompts marked with [? for suggestions]"*, and the sibling `prompt_with_suggestions` honours it. In `prompt_input` a `?` fell through to `echo "$result"`, and the caller's next line **saved it as the answer**.

Observed in a real intake: `one_time_budget` and `users_12mo` both stored as the literal string `"?"` and carried into `PROJECT_INTAKE.md`.

## The fix

Wrap the read in the loop the sibling already uses — read, test, `continue` — and emit a notice on **stderr**, so a caller capturing stdout gets a clean answer.

## Evidence

| | |
|---|---|
| GREEN | 9/0 |
| RED against a whole-file revert | 3/6 |
| python3 absent | 8/0, with S1 `[SKIP]`-ing loudly rather than passing |
| mutants | MP1 and MP2, each with a named killer |

Verified not to spin on EOF across nine input shapes under a watchdog.

## Disclosed

The entry widens your `## BUG-010:` defect (2) — an unbounded choice prompt on EOF — from `prompt_choice` to `prompt_with_suggestions`. That has since been confirmed **dynamically**, not just from source: called through command substitution as its call sites do, with no default, it spins at roughly 100 KB/s and needed SIGKILL. It is live at **12+ call sites** passing `""` as the third argument. Not fixed here; recorded as a residual.

Note for anyone reproducing that one: `</dev/null` does not help, because EOF is the trigger rather than the cure.


Asborien:
Force-pushed once to re-author the tip commit. A test lab had left a placeholder identity in my clone's config, so the last commit carried a wrong author and Signed-off-by. The tree is byte-identical; only the author and sign-off lines changed.

kraulerson:

PR #392, opened 2026-09-13
fix(upgrade): repair a manifest mode written in the wrong vocabulary

Asborien:
An adopted organizational project carries `mode: "organizational"` in its manifest, but the host drivers gate their org-tier rules on `mode = "org"`. The value matches neither arm of `jq -r '.mode // "personal"'`, so it passes through verbatim, matches no `org`-gated branch, and the org-tier assertions **never execute** — while the function returns 0 and the banner reports success.

## Measured against the real GitHub API

```
BEFORE   [OK] Ready: protection verified for organizational mode   exit 0   <- VACUOUS
AFTER    [OK] Ready: protection verified for org mode              exit 0   <- rules ran
```

A silent skip became a real check. The `gh api` call succeeded in both, so the second `[OK]` means `required_approving_review_count` and `required_status_checks` were genuinely asserted.

## The fix

`upgrade-project.sh:686` is, measurably, the **only writer of `.mode` in the harness**, so `_run_idempotent_backfill` is the only available repair path. The block repairs present-and-invalid only; an absent mode is left alone, and a manifest/phase-state disagreement is refused by name rather than guessed.

## Evidence

| | |
|---|---|
| GREEN | 15/0 |
| RED against the unfixed script | 6/9 |
| mutants | MP1–MP3, each with a named killer |

The `V` section proves the repair changes **behaviour**, not just the stored value: `V0` and `V2` are controls establishing that the driver's org rules both fire and can pass, so `V1` is the org rules running rather than a blanket failure.

Independently reproduced bit-for-bit on a real adopted project: running the entry point produced a manifest blob identical to the committed one, with every provenance field unchanged including `lastSyncDate`.

## Sequencing — please read

**This must ship before or with the mode-vocabulary refusal (separate PR).** That change makes an unknown mode a hard `return 1`. Without this backfill there is no repair path, because this block is the only writer of `.mode` — so shipping the refusal first strands every already-adopted organizational project.


kraulerson:

PR #393, opened 2026-09-13
fix(host-drivers): refuse an unknown mode instead of skipping the org rules

Asborien:
`host_verify_protection` accepted any string as `mode` and gated its org-tier rules on the literal `org`. A mode of `organizational` therefore matched nothing, the org-only assertions were skipped, and the function returned **0** — while the banner claimed verification.

```
mode=organizational, driver unguarded:
  [OK] Ready: protection verified for organizational mode      exit 0   <- assertions never ran
```

The project is told its org-tier branch protection is verified when only the personal-tier subset executed.

## The fix

All three drivers now refuse an unknown mode by name:

```sh
case "$mode" in
  personal|org) ;;
  *) echo "host_verify_protection: mode must be 'personal' or 'org', got '$mode'" >&2; return 1 ;;
esac
```

**This enforces a contract the file already stated twice** — `host-drivers/github.sh` documents `personal|org` at both `:112` and `:166`. It was written down twice and enforced once.

## Evidence

| | |
|---|---|
| GREEN | 33/0 |
| RED against unfixed | 13/20, the 13 survivors being the declared controls |
| mutants | MP1–MP6, each with a named killer |

`H2` is load-bearing as a control: it fails at `mode=org` naming an org-only rule, which proves the fixture *reaches* the org block — so `organizational` returning 0 is a skipped assertion rather than a fixture that never got there.

`init.sh` has never written `organizational` into `.mode`, so this cannot newly block a **scaffolded** project. It blocks an adopted one, which is the population the entry is about.

## Sequencing — this is the important part

**Ship the mode backfill (separate PR) first or together. Never this alone.**

Without the backfill there is no repair path: `upgrade-project.sh:686` is the only writer of `.mode` in the harness, and it is that PR. Shipping this refusal on its own converts a silent false pass into a hard refusal that affected projects **cannot clear**.

Both orders converge if both land; the hazard is shipping this one by itself.


kraulerson:

PR #394, opened 2026-09-13
fix(gates): resolve the hooks directory through git, not a literal path

Asborien:
`install-filesystem-gates.sh` resolved the hooks directory in a way that **refuses a valid linked worktree** and, separately, could write the BL-030 gate into a repository the caller never named.

## Two defects

**1. The repo guard tested the wrong thing.** `git rev-parse --is-inside-work-tree` answers "is this path *somewhere inside* a repo", not "is it a repo root". A plain directory nested inside a repository returns rc 0, and `--git-common-dir` then resolves relative to the **enclosing** repo. Reachable via `reconfigure-project.sh` on a project whose `.git` was removed.

**2. `[ -d "$PROJECT_ROOT/.git" ]` refuses a linked worktree**, where `.git` is a *file*. Proven in a real one:

```
OLD guard  -> FALSE -> '[FAIL] not a git repo', exit 1     <- refused a valid repo
NEW guard  -> --show-toplevel matches, ACCEPTED
             HOOKS_DIR resolves to the COMMON gitdir
```

For any project using worktree-per-branch, that meant **every worktree session ran with strict mode reporting green while the gate was absent and silent** — the emitted hook fell through with no output.

Also fixes two unguarded assignments under `set -euo pipefail`, where a git build rejecting `--path` killed the script with no diagnostic.

## Evidence

| | |
|---|---|
| GREEN | 16/0 |
| mutants | MU1–MU5, each with correct survivorship stated |

`R10` covers the refusal arm — a non-repo nested in a repo is refused and the enclosing repo's hooks dir is untouched. That arm previously had no coverage at all.

`R9` is source-level and now **discovers** the installer's call sites rather than iterating a hardcoded pair. A fixed list can only go stale in the permissive direction: a third caller added later would never be checked. It also made the case unrunnable in a project checkout, which ships `reconfigure-project.sh` but not `init.sh` — 15/1 forever, and a suite that can never go green teaches people to ignore it. Now: framework finds 2 and scores 16/0, a project finds 1 and scores 16/0, zero found still fails closed.

## Disclosed

`--uninstall` leaves `framework-gate.sh` in place and only strips the marker block. That is pre-existing and unchanged here, but the fix's own comment frames "leaving the marker block and framework-gate.sh behind" as part of the defect, which over-claims slightly what uninstall does.


Asborien:
Force-pushed once to re-author the R9 commit: a test lab had left a placeholder git identity in my clone's config, so it carried a wrong author and Signed-off-by. Its tree is byte-identical; only the author and sign-off lines changed.

Also added one follow-up commit on top, from an adversarial review of the R9 case as carried into a downstream project: discovery now keys on the installer's identity rather than the `bash "..." --install` spelling (a quoted-flag caller with a stderr swallow escaped the old regex once a second caller existed), the dead `missing` assignment is gone, and the pass line counts call lines. Identity-based discovery found a third real caller in `scripts/upgrade-project.sh` that swallowed the installer's stderr; it gets the same one-line change as the other two. 16/0 green with 4 call lines discovered; the quoted-flag mutant is 16/0 under the old suite and 15/1 under this one. That follow-up has not had an independent adversarial review.

kraulerson:

Two small additions from review: a bare `x="$(cd …)"` under `set -euo pipefail` aborted with zero output for a nonexistent root (now falls into your "not a git repo" arm, R11), and a `core.hooksPath` that resolves to the repo's own hooks dir is no longer refused (R12) — your message was right in every case but that one.

PR #395, opened 2026-09-13
fix(gates): drop two false statements from the self-approval remedy

Asborien:
The self-approval refusal in `check-phase-gate.sh` tells the operator two things, and **both are false**.

```
Governance requires a different individual to approve phase gates for organizational projects.
Have the approver commit the APPROVAL_LOG.md entry themselves, or use --force with documented justification.
```

**1. `--force` was never built.**

```
$ bash scripts/check-phase-gate.sh --force
[FAIL] Unknown argument: '--force'          exit 2
```

**2. "Have the approver commit the entry themselves" is the action that REPRODUCES the failure.** The check fails precisely when the commit author equals the Approver cell. Following the advice exactly guarantees the same refusal.

## The fix

Remove both. The replacement states only what the gate actually compared, admits what it cannot distinguish, and points a single-authority project at the pre-condition behind the symptom. **It asserts no rule**, so it cannot be invalidated by how the larger question below is answered.

## Evidence

| | |
|---|---|
| GREEN | 8/0 |
| RED against `ceb450e1` | 2/7 |
| mutants | MTR1 and MTR2, each reinstating one removed statement |

`R2` runs the flag and requires exit 2 with `Unknown argument`. That converts a one-off observation into a standing fact: **if anyone ever implements `--force`, R2 goes red and this removal gets revisited** rather than quietly becoming wrong.

## This is half the defect — the other half is yours to rule on

`check-phase-gate.sh:1719` fails when the commit author **equals** the approver. Three sources say the opposite:

| source | requires |
|---|---|
| `governance-framework.md` §V control 1 (`:180`) | *"MUST be committed ... by the **approver**, not the Orchestrator"* — author **=** approver |
| `governance-framework.md:183` audit rule | authors must **match** the listed approvers |
| baseline §5 invariant #9, quoted by the code at `:1502` | author **=** approver |

And `tests/test-check-phase-gate-self-approval.sh` T2 pins the code's side deliberately: *"approver committed by author → MUST FAIL"*.

The check never establishes who the Orchestrator **is**, so it cannot distinguish an independent approver signing their own row from an Orchestrator approving themselves. Control 3 forbids the latter; the implementation forbids both.

**No fix is proposed for that half**, because any wording of the form "do X and it will pass" has to pick one of two contradictory rules, and that is your call. The entry records three possible resolutions with no recommendation.


kraulerson:

One change: the shipped text's `See ## BL-274:` now reads `See ## BL-275:` — BL-274 was never filed on any branch, so it pointed at nothing; a placeholder entry now holds that number, and R7 pins the pointer.

PR #396, opened 2026-09-13
fix(reconfigure): read the pipeline host instead of inferring it

Asborien:
`reconfigure-project.sh --field language` regenerates the CI pipeline from a path built on a host value it never reads back, warns when that path does not exist, and then reports success anyway.

## Evidence

| | |
|---|---|
| GREEN | 14/0 |
| RED against `ceb450e1` | proven |

The fix reads the host through `host_read_from_manifest` rather than inferring it, so the regeneration either uses the project's declared host or refuses by name.

## Disclosed — a measured constraint, not a scoping preference

**The absent-host case cannot be completed from this script by any mechanism currently reachable from it.** `_detect_pipeline_host`, the framework's own remote-inference helper and the obvious thing to reuse, is defined only in `scripts/verify-install.sh:1319`. That file carries `set -euo pipefail` at `:2` and `guard_not_in_framework || exit 1` at `:20`, both at **top level with no sourced-detection guard**. Sourcing it does not import a function — it runs the verification script, and inside the framework repo it exits the caller.

There is no shared home to reach for either: the same four-arm inference is duplicated at **four shipped sites** — `check-gate.sh:159`, `upgrade-project.sh:511`, and `verify-install.sh` at both `:242` and `:1319` — while `scripts/lib/host.sh`, the shared host library, carries zero host-inference sites.

Hoisting it is a refactor, so nothing is proposed here. That observation is filed separately as the structural entry.


kraulerson:

Renumbered **BL-262 → BL-287** on the way in (number taken on `main` first). Recorded, not changed: the new `exit 1` paths leave `tool-preferences.json` updated with no rollback in scope — your diagnostics say so honestly; noted as a residual.

PR #397, opened 2026-09-13
fix(scout): stop claiming full-history scope on a shallow clone

Asborien:
Scout reports `scope: full-history` on a **shallow** clone, so a secrets scan that could only see one commit is recorded as having seen everything.

## Evidence

| | |
|---|---|
| GREEN | 13/0 |
| RED | proven against the unfixed scanner |

## The part worth reading — the first remedy was wrong, and the review caught it

The original fix told the operator, in five places, to run `git fetch --unshallow`. That advice is **wrong**, and wrong in exactly the way the defect itself is about.

`git-clone(1)`: **`--depth` implies `--single-branch`**. `--unshallow` deepens history without widening the refspec. So following the advice exactly leaves a credential sitting on an unfetched branch invisible — and Scout then reports `scope: full-history`, zero findings. The fix's own defect, reached by doing what the fix said.

Corrected remedy, verified to take findingCount 0 → 1:

```sh
git remote set-branches origin '*' && git fetch --unshallow
```

**This has since been exercised for real**: it is what unshallowed a contributor clone during this work, taking history depth from 3 to 1731 and making an identifier-collision sweep against merged history possible for the first time. The bare `--unshallow` would not have.

## Disclosed

A recorded flake travels with the entry: one Linux run at 12/1 against three at 13/0, uncaptured and unexplained. It is written down so the next person suspects the suite before the fix.


kraulerson:

Renumbered **BL-264 → BL-288**. Two additions: S9 located each remedy by a hardcoded line number and broke when an earlier change shifted one line — it now finds them by content; and S10 pins the half your fix promised but nothing tested (findings from the commits a depth-1 clone *can* read are emitted, not nulled) — the reviewer's narrowing mutant dropped two live credentials at 14/0 before it.

PR #398, opened 2026-09-13
docs(backlog): upgrading to organizational never writes .mode

Asborien:

`upgrade-project.sh --deployment organizational` never writes `.mode`.

The tier-change write is `'. + {deployment: $dep, poc_mode: $pm}'`; the BL-030 writes are `{deployment, poc_mode, enforcement_level}`. **None touches `.mode`.**

So a **scaffolded** project upgraded personal → organizational carries `deployment: organizational` with `mode: personal` permanently, and is verified against the personal bar for the rest of its life.

Neither the mode-vocabulary refusal nor the mode backfill catches it, because `personal` is a valid word to both — the backfill's guard is `case "$bl270_mode" in ''|personal|org) ;;`.

Same defect class as the vocabulary entry, on the scaffolded path rather than the adopted one.

---


kraulerson:

PR #399, opened 2026-09-13
docs(backlog): four reuse attempts, four different local reasons

Asborien:

Four separate attempts to reuse the framework's own helpers each failed, for four **different** local reasons.

| | why the reuse failed |
|---|---|
| `_bl145_*` helpers | `verify-install.sh` is not sourceable — `set -euo pipefail` at `:2` and `guard_not_in_framework` at `:20`, both top level |
| `_detect_pipeline_host` | same file, same reason |
| Scout's package-manager helper (CI) | Scout does not ship into projects |
| the same helper (release) | reachable, but nothing exists to detect at that moment — no lockfile at scaffold time |

Two mechanisms, one outcome: **not sourceable**, and **sourceable but not shipped**.

Measured: the framework carries **41** library files, a project receives **24**, and the 17 that do not ship are exactly `lib/adopt/` and `lib/scout/`.

Headline: `scripts/lib/host.sh` — the shared host library — carries **zero** host-inference sites, while four shipped scripts each carry the same four-arm `case`, `verify-install.sh` twice, about 1,080 lines apart.

The repair is a hoist, which is a refactor, so nothing is proposed. `# BL-095-STATE-READERS` is cited as the model — including that its own header names three deliberately unmigrated siblings with reasons.

---


kraulerson:

PR #400, opened 2026-09-13
docs(backlog): the bypass detector records other programs' output as Claude's

Asborien:

`scripts/hooks/bypass-detector.sh` extracts `.tool_response.{stdout,stderr,output,content}` on PostToolUse. The scanned surface is therefore **the command's output, not anything the model authored** — but a match writes a row with `actor: "claude"` and raises a **blocking** sentinel.

## Measured

- **6 of 61** rows are provably echoed tool output — four begin with a `path:NNN:` grep prefix, two with a `#` comment marker. No sentence proposing a bypass begins that way.
- **6** rows have an excerpt containing a nested `"excerpt":` field — the detector scanning its own ledger. One is **doubly** nested: the detector quoting its own quotation.
- **Three of those six are `CLAUDE.md` line 129** — the *Refuse-to-recommend* paragraph, which forbids proposing `--no-verify`, synthetic step completions and force-pushes. The rule tripped `no_verify`, `fake_loop` and `manual_step_complete` simultaneously. It lives in the governing document every agent is required to read end to end.
- Reading `tests/test-bypass-detector.sh:28` and `:57` — **your own committed test suite** — raises rows too.
- The ledger grew **61 → 64 during a few minutes of verifying this**. The count tracks how much documentation has been read, not how many bypasses were proposed.

## Why it is not merely noisy

Each false row raises a sentinel that **blocks commits**, and the sentinel is read from the session's project directory rather than the repo being committed to (filed separately), so it blocks work in unrelated repositories. Five were declined by hand in a single day, none corresponding to anything anyone proposed. That is `## BL-256:`'s unearned-receipt class running in reverse — an unearned **refusal**.

A correct regression suite for this cannot avoid containing the trigger strings, because asserting "the detector fires on a genuine proposal" requires a genuine-looking proposal in a fixture. **The defect obstructs its own remediation.**

## Why no fix is proposed

Every candidate narrows a security control you own; output-scanning is a tested contract (T1); and one option changes an audit-row schema and needs a sweep of `actor` readers. Three options are analysed in the entry with a recommendation, but choosing among them is a judgement about this framework's risk appetite rather than about correct code.

Lead on **provenance**, not pattern-matching: a comment-aware matcher would cut the rate and leave the class untouched, because a `cat` of any file containing the phrase would still do it.

---


kraulerson:

PR #401, opened 2026-09-13
docs(backlog): half the gate's warnings block and half do not

Asborien:

`[WARN]` and `[FAIL]` text in `check-phase-gate.sh` is cosmetic — the exit predicate is `if [ $issues -eq 0 ]`. `## BL-104:` already records this, and `CLAUDE.md` repeats it. **This entry does not re-file that.**

The measured finding is narrower: **34 of 71 `[WARN]` emit sites increment the blocking counter and 37 do not, and both print the identical label.** Nothing in the output, the exit code or the closing banner maps a label to a verdict.

Some warnings are honest about it — the Competency Matrix line says "WARN-first, not blocking" and genuinely does not increment. The pre-conditions warning at `:2014` says nothing about its own weight and blocks.

## Why it matters beyond naming

Isolating an exit-code assertion in a gate test required padding a fixture with `docs/phase-0/*`, a pre-conditions section and `PRODUCT_MANIFESTO.md` **purely to stop warnings counting**. Which warnings count is not discoverable from the output, so anyone writing a gate test has to diff the increment sites in the source — and their fixture silently rots when a WARN changes weight.

`## BL-256:` cuts both ways: a gate that refuses on evidence it presented as advisory is the mirror of one that passes on a check it never ran.

**34 is a floor** — the count only inspects the line following each emit site.

Three shapes sketched, none proposed: make the label carry the weight, split the counters, or lint the convention so it stops drifting.

---


kraulerson:

PR #402, opened 2026-09-13
docs(backlog): CI package-manager detection, held after measurement

Asborien:

Held after measurement, not chosen. Both halves are blocked by facts rather than by scope preference.

**The CI half** has four consumers, three of them project-side — and Scout, whose package-manager helper is the natural detector, does not ship into projects.

**The release half** has two consumers, one project-side. At the only moment the framework-side one runs, `init.sh` has written no `package.json` and no lockfile. So a lockfile detector returns nothing: it is not a resolution problem, **the information does not exist yet**.

Closing it needs a new CLI input, a new shipped library surface, or moving when the pipeline is rendered. All three are above the bar for a fix.

The withdrawn implementation's `SOIF_CI_PACKAGE_MANAGER` was its own invention — 1 occurrence on the branch, 0 on `main`.

**A hazard worth knowing regardless of this entry's fate:** CI templates carry zero placeholders today. If a second verbatim CI writer lands before any `__PM_*__` placeholders are introduced, `--field language` will write a pipeline full of literal placeholders over a working one, announced as `[OK] CI pipeline regenerated` — and neither change's tests would catch it, because the fixtures are synthetic templates that never contain a placeholder.

---


kraulerson:

Renumbered **BL-263 → BL-285** — your numbering paragraph predicted this exact collision, and it happened one number further along than you guessed. The original reasoning is kept as written under a "Numbering (as filed)" heading.

PR #403, opened 2026-09-13
docs(backlog): a single-branch clone is not shallow, same blind spot

Asborien:
**Backlog entry only — no fix is proposed.**

Filed separately from the shallow-scope entry at both reviewers' recommendation, rather than bundled.

A plain `--single-branch` clone is **not shallow**. `git rev-parse --is-shallow-repository` returns false and there is no `.git/shallow`, so Scout's shallow detection never fires — and the blind spot is **identical**: refs that were never fetched cannot be scanned, and the scan is nonetheless recorded as having covered the repository.

This is the same class as the shallow-scope defect (separate PR) approached from the other side. `--depth` implies `--single-branch`, which is why the two travel together — but the converse does not hold, and the narrowing detector only catches one of them.

---

Lints rc=0. Duplicate sweep done **by file touched** rather than by keyword.


kraulerson:

PR #405, opened 2026-09-13
fix(gates): read the pending-approval sentinel from the repo .cwd is in

Asborien:
`pa_check` in `scripts/pre-commit-gate.sh` reads the pending-approval sentinel as the bare relative literal `.claude/pending-approval.json`, and on the PreToolUse path the script never changes directory. So it resolves against the **session's** project directory, never against the repository the intercepted commit targets. There is no root resolution to be wrong; there is none at all.

Measured both ways on 2026-09-13:

- a sentinel in the session's project **blocked a commit in an unrelated clone** with no `.claude/` of its own;
- a scratch repository carrying **its own sentinel committed cleanly** from a session rooted elsewhere.

It gates every repository **except** the one it lives in.

## Two halves, one fixed

**Under-block — fixed here.** The Builder's Guide § "Structured Decision Points" ships this reader *into* each project (`upgrade-project.sh` "copies … the updated `scripts/pre-commit-gate.sh` into existing projects, so the enforcement … goes live immediately on upgrade") precisely so a project's own sentinel gates that project's own commits. It did not.

**Over-block — left open, deliberately.** A session sentinel halting commits in every other repository touched from that session. Both readings are defensible and the design states neither: the Builder's Guide model is per-project throughout, but the sentinel means "the agent is holding, do not advance", and advancing in a *different* repository is arguably the same failure the sentinel exists to stop. The entry records the measurement and the three options considered, and changes nothing about this half. The operational pain of it is real and belongs to `BL-277` (#400), which manufactures false sentinels from ordinary file reads; narrowing this control would relieve a symptom whose cause is elsewhere.

## Fix — additive, and narrower than the first cut claimed

`BL-278-SENTINEL-ROOT`: `_pa_target_sentinel` reads `.cwd` from the hook envelope, resolves it with `git -C "$c" rev-parse --show-toplevel`, and returns that repository's sentinel path. `pa_check` consults it **only when the session-relative read found nothing**, so nothing that blocks today stops blocking. Every unreadable or unresolvable input returns 1 and degrades to exactly today's behaviour.

**What it covers, precisely:** a bare `git commit` issued while `.cwd` is inside the sentinel's repository (including any subdirectory of it). `.cwd` is the directory Claude is in *before* the command runs, so the two other cross-repo shapes are **not** covered and are recorded as an open residual in the entry with the probes quoted:

| shape | `.cwd` | result |
|---|---|---|
| `git -C target commit` | session | allow — and `_is_git_commit` never matches `git -C x commit` at all (pre-existing) |
| `cd target && git commit` | session | allow |
| bare `git commit` | target (or below) | **deny on the target's sentinel** |

Resolving the target from the command text would close those, but it is a change to a security arm whose scope the design does not state, so it is left to you. C6 pins the current `cd &&` behaviour as a labelled residual so the case flips the day someone closes it.

Rejected: resolving *only* against the target repo (fixes both halves, but changes a security control whose intent is unstated — if you decide the per-project reading is right, that is a one-line follow-up and C2 is the case that changes); resolving from `CLAUDE_PROJECT_DIR` (names the session, which is the value already producing the wrong answer).

## Evidence

| | |
|---|---|
| GREEN at tip | `Results: 10 passed, 0 failed` rc 0 — bash 5.3 and `/bin/bash` 3.2.57 |
| RED against `main` | `Results: 5 passed, 5 failed` rc 1 — C1 and C1b fail for the right reason: rc 0 and **empty** output, the gate allowing silently |
| C1b | `.cwd` two levels below the target's root must still deny; this is the case that exercises the toplevel resolution |
| C4 | scoped to the sentinel decision: against the tip gate alone, a session sentinel denies and no sentinel gives no denial; against `origin/main` on identical fresh fixtures, the pending-approval deny envelope is identical or identically absent. Whole-output `cmp` was dropped because it would trip on any later gate output change and compare `main` with itself after merge |
| mutants | M1 removes the operative line → C1. M2 replaces the `rev-parse --show-toplevel` resolution with `root="$c"` → C1b alone, C1 still passing, which is the point. Both assert the operative text occurs exactly once before, zero after, one changed line, syntax-checked |
| lints | `lint-tests-registered`, `lint-backlog-references`, `lint-bl-markers`, `lint-no-live-remote-in-tests` all OK |

Three things the suite learned by control rather than review, recorded in the test commit: the fixture repos need an `origin` remote or an earlier arm of the gate denies first; assertions are scoped to the suite's own unique question strings, not "denied at all"; and the `origin/main` baseline must live in a complete mirror of `scripts/`, because the gate resolves siblings through `SCRIPT_DIR`.

Reproduce: `bash tests/test-bl278-sentinel-root.sh </dev/null` from a checkout at the tip.

## Disclosed

- **The first cut of this branch overstated the fix and had an untested line, both found by an independent adversarial review before submission.** The commit title and comment said "the repository the commit targets"; `.cwd` cannot provide that, as the table above shows. And every case set `.cwd` to the repo root, so replacing the toplevel resolution with `root="$c"` left the suite at 7/0. The comment, test header and entry are corrected in the second cut; the fix commit's title is immutable history and is wrong by the same margin.
- The first cut cited `docs/reference/builders-guide.md`, which does not exist here; the file is `docs/builders-guide.md`. Fixed, with bare line numbers replaced by section headings.
- With `.cwd` set to a **linked worktree** of a repository whose main checkout holds the sentinel, the gate allows. That is a property of the sentinel being an untracked per-checkout file, stated in the entry, not a defect.
- The 10-case suite was not re-run on Linux; the earlier 7-case Linux record stands labelled as such.
- A Linux container run (bash 5.2, ubuntu:24.04, non-root) also scored 7/0, but its C4 was **vacuous**: the container has no real upstream, so the baseline it compared against already carried the fix. Recorded in the entry rather than counted.
- `BL-277` is cited and lives on a sibling branch (#400); declared via `lint-backlog-references-ignore` rather than dropped.
- The fix commit shipped before the test commit on this branch; the TDD ordering gate warned and logged it on this tier. The suite was written and run RED against `main` before the fix was written; the commit order is the author's, not the work's.

Found while adopting the framework into a downstream project, out of the `BL-277` work. DCO-signed on every commit.


kraulerson:

PR #406, opened 2026-09-13
fix(test-gate): do not read an unmeasured bug source as zero bugs

Asborien:
`check_phase_gate()` in `scripts/test-gate.sh` counted open bugs from GitHub Issues whenever `gh auth status` succeeded, and took that as proof the repository tracks bugs there under the SEV vocabulary. It does not. `gh issue list --label SEV-1` on a repository with no such label returns an **empty list at rc 0**, and the gate reported "Phase 2→3 gate clear" having measured nothing.

## Two arms, one function

**Arm 1 — an absent label is indistinguishable from zero bugs, and the gate clears on it.** Measured on a 110-label repository carrying no `SEV-*` label:

```
gh issue list --label "SEV-1" --state open --json number | jq length   -> 0   rc 0
gh issue list --label "bug"   --state open --json number | jq length   -> 30  # control
```

Three routes in: the missing label; a repository with **no GitHub remote at all** (`gh` exits 1, empty stdout, and the existing count sanitizer turns that into 0); and **partial vocabulary** — the two SEV-2 arms query `--label SEV-2 --label fix-now` and `--label SEV-2 --label deferred`, two flags **intersect**, so a project that created `SEV-1..3` but not `fix-now`/`deferred` reads 0 from both **blocking** arms.

**Arm 2 — the counts saturate at 30.** No `--limit`, so every query stops at `gh issue list`'s default. Measured: `bug` → 30 without `--limit`, 257 with `--limit 1000`. **This arm does not let a gate pass** — every consumer is a `-gt 0` test, and saturation never yields 0 from a non-zero truth. It corrupts the number the operator triages against. Filed at its own severity, not arm 1's.

## Fix

| marker | |
|---|---|
| `BL-280-SEV-LABEL-PROBE` | one bounded `gh api repos/{owner}/{repo}/labels/<name>` per label, rc 0 present / rc 1 absent. **Not** `gh label list`, whose own default limit is 30 (measured) — that would rebuild arm 2 inside the fix for arm 1. |
| `BL-280-NO-SEV-VOCAB` | with none of `SEV-1/2/3` present, GitHub is no longer counted as a bug-tracking *source*; the function falls through to its **existing** honest arm (`[WARN] No bug tracking source found`, exit 2). No new policy minted. |
| `BL-280-PARTIAL-VOCAB` | vocabulary adopted but incomplete, GitHub the only source: the affected arm prints `NOT MEASURED … This is NOT a clean result` and sets `warnings=true` rather than `[OK]`. Gated on no `BUGS.md`, because a BUGS.md project counting 0 from GitHub is correct. |
| `BL-280-QUERY-LIMIT` | `--limit 1000` on all four queries. Residual stated: above 1000 open issues of one severity the figure is still understated; the verdict is not, because it is already a block. |

Blast radius: a `BUGS.md` project keeps all four `[OK]` lines and gains one `[INFO]`. A project with no `gh`, or not authenticated, is byte-identical. The only outcome that changes from clear to not-clear is the one that was never measured.

Two further markers landed after an adversarial review of the first cut (see *Disclosed*):

| marker | |
|---|---|
| `BL-280-QUERY-STATUS` | each `gh issue list` query's exit status is captured; a failed query marks that arm **not measurable** with the reason (`the 'SEV-1' query failed (gh exit 1)`) instead of sanitising rc≠0 to a count of 0 |
| `BL-280-REPO-PREFLIGHT` | one `gh repo view --json name` before the label probes; when `gh` cannot resolve a repository the note says so, rather than claiming no SEV label exists |

## Evidence

| | |
|---|---|
| GREEN at tip | `Passed: 19   Failed: 0` rc 0 |
| RED against `main` | `Passed: 3   Failed: 16` rc 1 — the three passes are honest-outcome controls (P0 pins the shim as the `gh` the gate calls; R0 a clean project; R7 a partly-adopted vocabulary with `BUGS.md`) |
| mutants | MT1 strips `--limit` → R4; MT2 reinstates `has_bugs=true` on the no-vocabulary branch → R1; MT3 drops the `BUGS.md` guard → R7; MT4 drives R6's predicate against a stripped file, 4/4 detected, 0 false positives on the real one; MT5 and MT6 delete the single-severity PARTIAL-VOCAB lines → R3b, R3c; MT7 discards the query status → R8. Each compares a cksum taken before the edit so a mutation that lands nothing fails loudly. |
| prerequisite | fails **closed** without `jq`: `[FAIL] PREREQ`, exit 1 — measured against a PATH holding one symlink, after a first attempt with `PATH=/usr/bin:/bin` proved fake because `/usr/bin/jq` exists on this host |
| lints | `lint-tests-registered`, `lint-backlog-references`, `lint-bl-markers`, `lint-no-live-remote-in-tests` all OK; shellcheck 0.11.0 clean |

Reproduce: `bash tests/test-bl280-bug-gate-unmeasured-source.sh </dev/null` from a checkout at the tip.

## Two judgements left to you, deliberately

1. **Should an unmeasurable bug source BLOCK rather than warn?** This fix routes it to exit 2 (attestation), matching the function's existing no-source arm. `BL-112-SAST-NOTRUN` argues at length that a gate people cannot pass is a gate they work around. That is a security posture, not an implementation detail.
2. **The gate has no record of which tracker the project uses, and could have one.** `build-progress.json`'s `bug_tracker` field is written in two places, both hardcoded to `"github_issues"`, and read by nothing; the intake wizard's own answer is saved elsewhere and consulted by nothing. `BL-203-INTERVAL-PLUMB` one field over. Proposed in the entry, not built, not filed.

## Disclosed

- **The first cut of this branch had two holes, found by an independent adversarial review before submission and fixed in the second cut.** (1) The two single-severity `PARTIAL-VOCAB` lines (SEV-1, SEV-3) had no case exercising them: deleting either left the suite at 13/0. R3b, R3c, MT5 and MT6 close that. (2) A query that failed *after* a successful label probe was still sanitised to 0 and printed four `[OK]` all-clears, and R2's pass string claimed otherwise. `BL-280-QUERY-STATUS`, R8 and MT7 close that, and R2's string now says what it proves. The entry records both under its "an earlier cut of this entry said" convention.
- One behaviour change from that second cut worth your eye: a `BUGS.md` project whose labelled GitHub query fails now gets a NOT MEASURED warning for that arm instead of `[OK]`. Recorded in the entry's blast-radius paragraph.
- Arm 2 was reported to the author as failing permissively in the same direction as arm 1. The entry corrects that framing after reading all four consumers.
- Two defects in the suite itself were caught by running it and are recorded in the test commit: R6's predicate was comment-blind, and MT1/MT2 originally "landed" against `main` while mutating nothing. That is why the RED tally moved from 5/8 to 3/10.
- Two BL tokens cited (`BL-275`, `BL-279`) are real entries on sibling branches (#395, #401) and resolve once those land; declared via `lint-backlog-references-ignore` rather than dropped.
- `## BL-280:` cites two `## BL-2xx:` entries that do not exist on `main` yet (see above); everything else resolves on this base.

Found while adopting the framework into a downstream project. DCO-signed on every commit.


kraulerson:

One addition: an enriched fixture makes the exit code a real assertion — R12 forces a failed `gh issue list` and requires rc 2. Your header explained exactly why rc could not be asserted on a bare fixture, and that reasoning was correct; a richer one closes it.

PR #407, opened 2026-09-14
docs(backlog): BL-282, no route to correct a recorded intake answer

Asborien:
Entry only, by decision. Once a section is in `completed_sections` the wizard will not ask it again, and nothing else offers to: `--resume` skips it by membership test, `reconfigure-project.sh --field` knows seven keys of the 122 the wizard records, and the wizard's flag surface has no generic writer.

## Measured

A downstream adoption's `intake-progress.json` holds `monthly_budget` as the literal string `"3"`. The operator typed `?` at the prompt, was shown a numbered list, and typed the number of the one they meant; `prompt_with_suggestions` has no numbered-selection arm. Reproduced on `main` with a canned section 3 (transcript in the entry), then the three closed doors:

```
--resume            -> [INFO] Resuming from Section 5   (lines mentioning Section 3: 0)
reconfigure --field monthly_budget --old 3 --new '$50-500/month'
                    -> [FAIL] Unknown field: monthly_budget   exit 1
```

## Options, not decided here

1. `--reask N` on the wizard.
2. `--set KEY VALUE` on the wizard, then re-render.
3. Extend `reconfigure-project.sh --field`, which already carries the approval-log audit row.
4. Document hand-editing `intake-progress.json` as the supported route.
5. Narrower: accept a number at `prompt_with_suggestions`.

Every one adds or documents a CLI surface you will own, and a documented surface is the hard-to-reverse kind. The contribution is the measurement and the options.

## Aside

`reconfigure-project.sh`'s refusal lists `track` and `deployment` as Supported while its `--help` says they are not supported there.

Lints clean: `lint-bl-markers`, `lint-backlog-references`, `lint-tests-registered`, `lint-no-live-remote-in-tests`. DCO-signed.


kraulerson:

Asborien:
Measured consequence of BL-282 on an adopted project, 2026-09-17: the intake had stored the literal "3" for monthly_budget (the operator meant the third suggestion). With no correction route in the wizard and monthly_budget outside reconfigure-project.sh's seven supported fields, the project corrected .claude/intake-progress.json and PROJECT_INTAKE.md by hand, recorded as an out-of-band data fix in the commit. A --set-answer <key> <value> on the wizard, or adding intake answers to reconfigure-project.sh's field list, would have made that a harness action rather than a hand edit.

PR #408, opened 2026-09-14
docs(backlog): BL-283, prompt_yes_no answers itself under harness mode

Asborien:
Entry only; may be by design. `prompt_yes_no` in `scripts/lib/helpers-core.sh` returns 1 without reading stdin under `SOIF_NONINTERACTIVE`, `CI`, or a non-TTY, printing a `[WARN]` that it defaulted to N. The wizard overrides `prompt_input` with an unconditional read but inherits `prompt_yes_no`, so a harness-driven intake, which the wizard's own refusal text invites ("set SOIF_NONINTERACTIVE=1 to drive it from a harness"), records "ZDR not attested" whatever the canned line says, and that line lands on the next prompt, the exception reason.

## Measured on `main`

Canned section 5, classification `internal`, then `Y`, then `pause`:

```
[WARN] Non-interactive context: skipping prompt ("Is ZDR ... [Y/n]") — defaulting to 'N'
[OK]   Phase 1 artifacts persisted (classification=internal, zdr_attested=false)
zdr_attested = 'false'   zdr_attestation_reason = 'Y'
```

Same with a plain pipe and no `SOIF_NONINTERACTIVE`. Control with classification `public`: no warning, empty reason.

**Consequence:** `check-phase-gate.sh`'s ZDR predicate is attested **or** a non-empty reason, so the Phase 1→2 ZDR gate clears on a one-character exception nobody wrote.

## Two readings

- **Correct:** a yes/no must not be answerable by canned input. Defensible for the helper's other ten callers.
- **Inconsistent:** every other wizard prompt reads canned input; this one alone does not, and the stream shifts by one from there.

Three shapes offered: a wizard-local `prompt_yes_no` (recommended, since the inconsistency is the wizard's); leave it and document it beside the refusal text; or make the shared helper read stdin under the explicit `SOIF_NONINTERACTIVE` opt-in only. The call is yours: whether a harness should be able to attest ZDR at all.

Lints clean: `lint-bl-markers`, `lint-backlog-references`, `lint-tests-registered`, `lint-no-live-remote-in-tests`. DCO-signed.


kraulerson:

PR #409, opened 2026-09-14
fix(intake): resume by position in the section list, not by number plus one

Asborien:
`scripts/intake-wizard.sh --resume` after a clean completion of section 11.5 runs no further sections and prints "Intake Complete!", skipping 12 and 13. Section 13 is the Agent Initialization Prompt Phase 0 needs.

## Mechanism

The runner skips with `[ "$section" -lt "$start_section" ]` over a list that is not monotonic, `(1 2 3 4 5 6 7 8 9 10 11 115 12 13)`, and `--resume` starts at `LAST_SECTION + 1`, which after 11.5 is 116. Nothing in the list is 116 or greater, so every section is skipped and the completion banner prints.

Reproduced on `main` with a progress file at `last_section: 115`:

```
[INFO] Sections completed: 1 2 3 4 5 6 7 8 9 10 11 115
[INFO] Resuming from Section 116
║              Intake Complete!                           ║
exit=0    last_section = 115
```

No `[STEP] Section 12` or `13`. `BL-266` (#390) guards the pause path only and does not reach this: it fires when 11.5 completes cleanly and the process then stops.

## Fix

| marker | |
|---|---|
| `BL-281-SECTION-ORDER` | the ordered section list defined once |
| `BL-281-NEXT-SECTION` | the successor computed by position in that list |
| `BL-281-POSITION-SKIP` | the runner skips on position, not on integer value |
| `BL-281-RESUME-POINT` | both resume paths route through the helper |

Same fixture on the fix: `Resuming from Section 12`, sections 12 and 13 run, `last_section` 13, banner after 13.

## Evidence

| | |
|---|---|
| GREEN | `Results: 19 passed, 0 failed` rc 0, under `/bin/bash` 3.2.57 and bash 5.2 in `ubuntu:24.04` as a non-root user |
| RED against `main`'s wizard | `Results: 3 passed, 16 failed` rc 1; C1 to C3 are controls that pass at base; R1 fails on `Section 12 never ran: [INFO] Resuming from Section 116` |
| mutants | MP1 puts the resume point back to last + 1, R1 stops it; MP2 drops the reached flag so 12 runs and 13 is skipped, R2 stops it. Each asserts the mutation landed before concluding |
| neighbours | `tests/test-brownfield-wp9-act-boundaries.sh` (unit lane) 29 / 0 on this branch; `tests/edge-cases-scripts.sh` E21, the resume case, passes on this branch. Its E18 fails on this machine with `upgrade-project.sh: cd: null directory`, a file this branch does not touch |
| lints | `lint-tests-registered`, `lint-backlog-references`, `lint-bl-markers`, `lint-no-live-remote-in-tests` all OK |

Reproduce: `bash tests/test-bl281-resume-after-115.sh </dev/null` from a checkout at the tip.

## Disclosed

- The interactive "Resume or start over?" menu is verified through the helper cases and by reading, not driven end to end, because it is TTY-only.
- Found by a downstream adoption that ran its intake with this hazard ahead of it. DCO-signed on every commit.


kraulerson:

